PR #3, opened 2018-05-16
First iteration of assignment 1

michbl:


michbl/assignment1/assignment1.cpp
```diff
@@ -0,0 +1,104 @@
...
+    int size = l.size();
+    if (k > size - 1)
+        throw runtime_error("k is too big");
+    int idx = size - k - 1;
+    typename list<T>::iterator it = l.begin();
+    for (int i = 0; i < idx; i++) {
+        it++;
+    }
+    return *it;
+}
```
Brusi: Good solution, well done :)
You used C++ standard list, which is perfectly fine.
As an 'extra credit' assignment, if you want, you can write your own simple implementation of a linked list and use it in your solution.

michbl/assignment1/assignment1.cpp
```diff
@@ -7,84 +7,129 @@ using namespace std;
...
+
+// Check whether a character is an english letter
 bool isLetter(char c) {
-    return (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z');
+    return isUpperCase(c) || isLowerCase(c);
+}
+
+char toUpperCase(char c) {
+    if (!isLetter(c)) {
+        return 0;
```
Brusi: This is a little unexpected... Why turn a non-letter to zero?
If this is an 'error', I suggest you do the following:
1. Add a comment on this function which clarifies this, e.g.
`// Returns zero if input character is not a letter`
(because the function name does not reflect that).
2. Whenever you call this function and you did not precheck that the input is a letter, check that its output is not zero (otherwise this function is not safe to use). I see that in your code there are no such occurances.

Another option is to make it return the same character in case input is not a letter. I think it makes this function easier to use (for example, you can turn an entire sentence to uppercase by calling this function on each character, without checking for spaces and punctuation).

michbl/assignment1/assignment1.cpp
```diff
@@ -7,84 +7,129 @@ using namespace std;
...
+    }
+    words.push_back(sentence.substr(pos));
+    return words;
+}
+
+// Returns an iterator to a word in wordsList which is an anagram of wordToFind.
+// Returns wordsList.end() if no anagram found.
+vector<string>::const_iterator findAnagramInList (const vector<string>& wordsList, const string& wordToFind, bool caseSensitive) {
+    for (vector<string>::const_iterator it = wordsList.begin(); it != wordsList.end(); it++) {
+        if (areWordsAnagrams(*it, wordToFind, caseSensitive)) {
```
Brusi: Good function reuse :)

michbl/assignment1/assignment1.cpp
```diff
@@ -7,84 +7,129 @@ using namespace std;
...
-            }
-            it2++;
-        }
-        if (!match)
+        vector<string>::const_iterator it2 = findAnagramInList(words2, *it1, caseSensitive);
+        if (it2 == words2.end()) {
             return false;
+        }
+        else {
+            words1.erase(it1);
```
Brusi: Questions for thought:
What is the complexity of this algorithm? Can you do better?

michbl/assignment1/assignment1.cpp
```diff
@@ -93,8 +138,8 @@ bool anagrams(string str1, string str2, bool caseSensitive) {
 template <typename T>
 T kthToLast(list<T> l, unsigned int k) {
     int size = l.size();
-    if (k > size - 1)
-        throw runtime_error("k is too big");
+    if (k < 0 || k >= size)
```
Brusi: `k` is still an unsigned int, so checking `k < 0` always returns false.

michbl:
- I agree with your comment about toUpper/LowerCase, I would make it return c in case c is not a letter.
- Forgot to delete the 'unsigned' from k's definition in kToLast..
- About the complexity of the anagrams algorithm, I'm pretty sure I can do better if I sort the words and the sentences, but I've decided to merge the current solution due to lack of time :(
Thanks for the feedback, promise to do better next time!

Brusi:
Just to clarify:
I was not implying that the complexity is not good enough, only that it could be interesting to analyze it and think of other ways to solve this in terms of efficiency, even without writing anything down :)

Complexity is part of the challenge if you choose to take it, but code readability, reusability and cohesion are as important, and this is what I chose to focus on in this assignment.

PR #5, opened 2018-05-21
Assignment 1 by Olga

July-Morning:
First iteration of the first assigment.
--
Olga Chistiakova


July-Morning/assignment1/Assignment_1_Q1.cpp
```diff
@@ -0,0 +1,91 @@
...
+// true - if strings are anagrams 
+// false - if strings are not anagrams or at least one string is empty 
+bool Is_Anagram (std::string s1, std::string s2)
+{
+    if ((s1.size() == 0) || (s2.size() == 0))
+        return false;
+    if (s1.size() != s2.size())
+        return false;
+    std::unordered_map<char, int> symbol_counter;
+    for (auto& sym : s1)
```
Brusi: A string is a sequence of characters, so a foreach loop iterates on type `char`.
`char` is a primitive, so there is no need to use a reference to it.
(A reference is implemented by a pointer; which means that any further access to the character will dereference the pointer, making the code less efficient than using the primitive directly).

To use the primitive directly:
`for (char sym : s1)`
or, if you insist on using `auto`:
`for (auto sym : s1)` (without &).

There are many opinions on using `auto`. I think that for primitives it is best to use the actual type name and not hide it behind `auto`, so I am in favor or using `char` here, which will be easily identified by experienced C/C++ users.

July-Morning/assignment1/Assignment_1_Q1.cpp
```diff
@@ -0,0 +1,91 @@
...
+        symbol_counter[sym]++;    
+    for (auto& sym : s2)
+        symbol_counter[sym]--;    
+    for (auto& sym : s2)
```
Brusi: That's a nice trick: going only over one of the strings characters is enough to check anagrams.
Another option is to go over the entries on the symbol counter:
```C++
for (const pair<char, int>& entry : symbol_counter) {
    if (entry.second != 0)
        return false;
}
```
This way, if the strings are very long (e.g. `str1 == str2 == aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa`), this loop goes only once per different character.
Also you don't need to look up every character in the map again (map lookup is O(log n) for std::map, and requires hash function to run in std::unordered_map. Not very expensive but still more than nothing.)
This does not improve the 'worst case' complexity but still may save some CPU cycles :)

P.S.
This loop is a good candidate to use `auto`:
```C++
for (const auto& entry : symbol_counter) {
    if (entry.second != 0)
        return false;
}
```

July-Morning/assignment1/Assignment_1_Q1.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+// Check if two strings are anagrams (case insensitive algorithm)
+// Return value:
+// true - if strings are anagrams 
+// false - if strings are not anagrams or at least one string is empty
+bool Is_Case_Insensitive_Anagram (std::string s1, std::string s2)
+{
+    std::transform(s1.begin(), s1.end(), s1.begin(), ::tolower);
+    std::transform(s2.begin(), s2.end(), s2.begin(), ::tolower);
+    return (Is_Anagram(s1, s2));
```
Brusi: No need for outer parenthesis:
`return Is_Anagram(s1, s2);`

July-Morning/assignment1/Assignment_1_Q1.cpp
```diff
@@ -0,0 +1,91 @@
...
+        if (symbol_counter[sym] != 0)
+            return false;
+    return true;        
+}
+
+// Check if two strings are anagrams (case insensitive algorithm)
+// Return value:
+// true - if strings are anagrams 
+// false - if strings are not anagrams or at least one string is empty
+bool Is_Case_Insensitive_Anagram (std::string s1, std::string s2)
```
Brusi: Good reuse of `Is_Anagram` from above :)

July-Morning/assignment1/Assignment_1_Q2.cpp
```diff
@@ -0,0 +1,104 @@
+#include <iostream>
+
+typedef struct node
```
Brusi: In C++ you can write structs using the same style as class, no need for old C style with typedef:

```C++
struct node {
    int data;
    node* next;
};
```

July-Morning/assignment1/Assignment_1_Q2.cpp
```diff
@@ -0,0 +1,104 @@
...
+    node * next;
+} node;
+
+class single_linked_list 
+{
+    private:
+        struct node * head;
+        
+    public:
+        single_linked_list()
```
Brusi: You can use an constructor initalizer list:
```C++
single_linked_list() : head(NULL) {}
```

July-Morning/assignment1/Assignment_1_Q2.cpp
```diff
@@ -0,0 +1,104 @@
...
+            head = NULL;
+        }
+        
+        ~single_linked_list() 
+        {
+            while (head != NULL)
+            {
+                node * temp = head; 
+                head = head->next;
+                delete head;
```
Brusi: DID YOU MEAN `delete temp;` ?

(Sorry, old habit)

July-Morning/assignment1/Assignment_1_Q2.cpp
```diff
@@ -0,0 +1,104 @@
...
+        // Find k-th element from the end of the list
+        // Return value:
+        // true - if k-th element from the end was found successfully
+        // false - if size of the list is less than given k
+        bool find_k_from_tail (int k, int& data) 
+        {
+            node * tail = head;
+            node * k_node = head;
+            int diff_counter = 0;
+            while (tail != NULL)
```
Brusi: I understand that first you advance `diff_counter` until it gets to (k+1);
then you advance k_node.

This can be better expressed by breaking this to two loops:
```C++
while (tail != NULL && diff_counter < k + 1) {
    tail = tail->next;
    diff_counter++;
}
while (tail != NULL) {
    tail = tail->next;
    k_node = k_node->next;
}
```
This code may be longer but easier to follow.
It is easier to understand each loops purpuse:
First loop: Advance `tail` `k+1` times.
Second loop: Advance `tail` and `k_node` together until `tail` reaches the end.

July-Morning/assignment1/Assignment_1_Q2.cpp
```diff
@@ -0,0 +1,104 @@
...
+    std::cout << " **** General tests for correctness **** " << std::endl;
+    test_list->insert(20);
+    test_list->insert(25);
+    test_list->insert(30);
+    Test_K_Search (test_list, 0);
+    Test_K_Search (test_list, 1);
+    Test_K_Search (test_list, 2);
+    Test_K_Search (test_list, 3);
+    Test_K_Search (test_list, 4);
+    Test_K_Search (test_list, 5);
```
Brusi: What about:
```C++
Test_K_Search (test_list, -1);
Test_K_Search (test_list, -100);
```
Does your code handle those? ;)

July-Morning/assignment1/Assignment_1_Q2.cpp
```diff
@@ -41,64 +39,94 @@ class single_linked_list
             {
                 temp->next = head;
                 head = temp;
-            }    
+            }
+            return;
```
Brusi: No need to write `return` at the end of a `void` function.

July-Morning/assignment1/Assignment_1_Q2.cpp
```diff
@@ -41,64 +39,94 @@ class single_linked_list
...
+template <typename T> void Test_K_Search (single_linked_list<T> * list, int k)
 {
-    int k_list_el;
+    T k_list_el;
     if (list->find_k_from_tail(k, k_list_el))
         std::cout << k <<"-th list element from tail is " << k_list_el << std::endl;
     else
-        std::cout << "Invalid input: list size is less than given k!" << std::endl;
+        std::cout << "Invalid input!" << std::endl;
     return;
```
Brusi: Same here (no need to `return`).

PR #6, opened 2018-05-21
Annnufan 1

annnufan:
Add solutions for Q1 and Q2


annnufan/assignment1/solution_1.cpp
```diff
@@ -0,0 +1,104 @@
...
+#include <string>
+#include <cstring>
+#include <vector>
+#include <algorithm>
+#include <cctype>
+#include <functional>
+
+void word_anagram(bool case_sens) {
+	std::string word_1, word_2;
+	std::cin >> word_1 >> word_2;
```
Brusi: You should write your code in a way that it is portable and ready to be used as part of a larger system.
Think of anyone trying to use this code in a different context:
What if the input is not from the keyboard, but from a file, from an internet database or any other source?
What if they want to use the return value as part of a backend system, and not only print it to the screen?
In that case, the best way to expose this code is using a function, for example:
```C++
bool word_anagram(string word_1, string word_2, bool case_sens);
```
(The return value can hold whether the input words are anagrams or not).

For this reason, always try to separate the user input/output and the actual assignment logic.
For example, you can wrap this function later with the user input and output:
```C++
std::string word_1, word_2;
std::cin >> word_1 >> word_2;
if (word_anagram(word_1, word_2, case_sens)) {
  cout << "anagrams";
} else {
  cout << "not anagrams";
}

File: annnufan/assignment1/solution_1.cpp [hunk 2]
```diff
@@ -0,0 +1,104 @@
...
+
+	std::vector<int> count(256, 0);
+	for (auto ch : word_1) {
+		count[ch]++;
+	}
+	for (auto ch : word_2) {
+		if (count[ch] > 0) {
+			count[ch]--;
+		} else {
+			std::cout << "Second word isn't anagram of first\n";
```
Brusi: Line 29 is a duplication of line 17.
This can be avoided when separating the in/out and the logic.

annnufan/assignment1/solution_1.cpp
```diff
@@ -0,0 +1,104 @@
...
+			return;
+		}
+	}
+
+	std::cout << "Yes, second word is anagram of first\n";
+	return;
+}
+
+std::vector<std::string> line_to_words(std::string line) {
```
Brusi: This function does two things:
(1) Splits lines to words.
(2) Sorts each word.

(2) is not reflected in the functions name, and a of this function can be surprised that the words are sorted.
What can we do?

We can add comments to reflect that:
```C++
// Each word in the result vector is alphabetically sorted.
```
Better than a comment, we can choose a name that reflects that:
```C++
std::vector<std::string> line_to_sorted_words(std::string line) {
```
But even better is to make this function do only one thing and name it accordingly.
This function can *only* split a line to words, and only later you can go over the result and sort each word.
This way this function can be better reused by a different user which only wants to split a sentence (very common use) without sorting the words (which is not very common).

annnufan/assignment1/solution_1.cpp
```diff
@@ -0,0 +1,104 @@
...
+	return words;
+}
+
+void sentence_anagram(bool case_sens) {
+	std::string line_1, line_2;
+	std::getline(std::cin, line_1);
+	std::getline(std::cin, line_1);
+	std::getline(std::cin, line_2);
+	if (!case_sens) {
+		std::transform(line_1.begin(), line_1.end(), line_1.begin(), std::ptr_fun <int, int>(std::tolower));
```
Brusi: No need to `ptr_fun` here, std::transform works with a template predicate, meaning you can pass anything with a matching `operator()` including native functions.
```C++
std::transform(line_1.begin(), line_1.end(), line_1.begin(), std::tolower);
```

Also, `ptr_fun` is deprecated according to [cppreference](http://en.cppreference.com/w/cpp/utility/functional/ptr_fun).

annnufan/assignment1/solution_1.cpp
```diff
@@ -0,0 +1,104 @@
...
+	if (word != "") {
+		std::sort(word.begin(), word.end());
+		words.push_back(word);
+	}
+	return words;
+}
+
+void sentence_anagram(bool case_sens) {
+	std::string line_1, line_2;
+	std::getline(std::cin, line_1);
```
Brusi: Please separate the input/output than the actual logic, here too.

annnufan/assignment1/solution_1.cpp
```diff
@@ -0,0 +1,104 @@
...
+}
+
+int main() {
+	int type_of_ask, case_sens;
+	std::cout << "I need information about type of question: is it only one word or whole sentence?\n1 -- one word\n2 -- whole sentence\n";
+	std::cin >> type_of_ask;
+	std::cout << "Do you want case sensitive or case insensitive anagrams?\n1 -- case sensitive\n2 -- case insensitive\n";
+	std::cin >> case_sens;
+	if (type_of_ask == 1) {
+		word_anagram((case_sens == 1)?true:false);
```
Brusi: `(case_sens == 1)?true:false` has the same logic as `case_sens == 1`, you can make it shorter if you want.

annnufan/assignment1/solution_2.cpp
```diff
@@ -0,0 +1,65 @@
+#include <string>
+#include <iostream>
+
+template <typename T>
+class node;
```
Brusi: I don't think there is a need to declare `node` here; a class can have a reference to itself without pre-declaring it.

annnufan/assignment1/solution_2.cpp
```diff
@@ -0,0 +1,65 @@
...
+		next = nxt;
+	}
+
+};
+
+template <typename T>
+node<T>* find_kth_to_last(node<T>* head, int k) {
+	node<T>* tail = head;
+	node<T>* kth = head;
+	int between = 0;
```
Brusi: `tail` and `kth` are good names, but the name `between` does not help understand the purpose of this variable.
Maybe `count` is better?

Brusi: `tail` and `kth` are good names, but the name `between` does not help understand the purpose of this variable.
Maybe `count` is better?

annnufan/assignment1/solution_2.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+	void set_next(node<T>* nxt) {
+		next = nxt;
+	}
+
+};
+
+template <typename T>
+node<T>* find_kth_to_last(node<T>* head, int k) {
+	node<T>* tail = head;
```
Brusi: `tail` and `kth` are good names, but the name `between` does not help understand the purpose of this variable.
Maybe `count` is a better name for it?

annnufan/assignment1/solution_2.cpp
```diff
@@ -0,0 +1,65 @@
+#include <string>
+#include <iostream>
+
+template <typename T>
+class node;
+
+
+template <typename T>
+class node {
```
Brusi: Nice encapsulate node implementation :)

annnufan/assignment1/solution_1.cpp
annnufan: If I just write ```std::transform(word_1.begin(), word_1.end(), word_1.begin(), std::tolower);``` it isn't compile.

annnufan:
I think I'm ready to next Iretation

annnufan/assignment1/solution_1.cpp
```diff
@@ -0,0 +1,118 @@
...
+bool word_anagram(std::string word_1, std::string word_2, bool case_sens) {
+	if (!case_sens) {
+		std::transform(word_1.begin(), word_1.end(), word_1.begin(), ::tolower);
+		std::transform(word_2.begin(), word_2.end(), word_2.begin(), ::tolower);
+	}
+	if (word_1.length() != word_2.length()) {
+		return false;
+	}
+
+	std::vector<int> count(256, 0);
```
Brusi: The number range of `char` type is -127 to 128.
So you can use an `unsigned char` instead (which range is 0-255).
Another option is to assume for the scope of this assignment that "extended char" values are not used (i.e. only 0-127 values are possible). In that case, `count` vector can be of size 128 instead.
(Either way, signed char can never hold value of 128 or more).

annnufan/assignment1/solution_1.cpp
```diff
@@ -0,0 +1,118 @@
...
+		if (count[ch] > 0) {
+			count[ch]--;
+		} else {
+			return false;
+		}
+	}
+	return true;
+}
+
+void solve_for_word(bool case_sens) {
```
Brusi: Nice helper functions that handle all the I/O.
They are short, simple and easy to understand, easy to use whenever you want, and keep the logic separated so you can use the logic without the I/O. Good work!

annnufan/assignment1/solution_1.cpp
```diff
@@ -0,0 +1,118 @@
...
+bool sentence_anagram(std::string line_1, std::string line_2, bool case_sens) {
+	if (!case_sens) {
+		std::transform(line_1.begin(), line_1.end(), line_1.begin(), std::ptr_fun <int, int>(std::tolower));
+		std::transform(line_2.begin(), line_2.end(), line_2.begin(), std::ptr_fun <int, int>(std::tolower));
+	}
+	if (line_1.length() != line_2.length()) {
+		return false;
+	}
+	std::vector<std::string> words_1, words_2;
+	words_1 = sorted_words(line_to_words(line_1));
```
Brusi: Thanks to separating to functions with meaningful names, this line is more readable.
The reader knows that they get an array of sorted of words.

annnufan/assignment1/solution_1.cpp
```diff
@@ -0,0 +1,118 @@
...
+	words_2 = sorted_words(line_to_words(line_2));
+	if (words_1.size() != words_2.size()) {
+		return false;
+	}
+	std::sort(words_1.begin(), words_1.end());
+	std::sort(words_2.begin(), words_2.end());
+	for (int i = 0; i < (int)words_1.size(); i++) {
```
Brusi: You can compare the vectors directly and get the same effect:
```C++
return words_1 == words_2;
```

If that was not possible, you could extract this loop to a function to get a more concise function.
(e.g. end the function with `return vectors_are_equal(words_1, words_2);`)

annnufan/assignment1/solution_1.cpp
annnufan: Could I add 127 for all  char before counting? I think it's be solve of problem

annnufan:
I correct ambiguous cases.

annnufan/assignment1/solution_1.cpp
Brusi: You can avoid the +127 in this case: You can use the implicit conversion from `char` to `unsigned char` to do the work for you. It is also more safe (an unsigned char can never be negative!).

So you can explicitly state the type in the loop:
```C++
for (unsigned char ch : word_1) { ... }
```
This way each character will be converted to 0-255 range and accessing the array is safe.

annnufan/assignment1/solution_1.cpp
```diff
@@ -16,12 +16,14 @@ bool word_anagram(std::string word_1, std::string word_2, bool case_sens) {
 	}
 
 	std::vector<int> count(256, 0);
+	
+	//String used signed char. So, it needs to add 127 for each char for counting sort
```
Brusi: There are many approaches of using `auto`. In my opinion, we should use it when it makes the code shorter but does not "hide" information from the user.
In my opinion, using `char` (or `unsigned char`) explicitly (as in `for (char ch : word_1)`) is more readable, especially if the reader is not sure whether it is signed or unsigned. (In general I think that `auto` is not a good replacement for primitive types).

Also:
string operator[] does not return a char, but a char reference! (`char&`)
This means that if you change `ch` in the loop, it will modify the string. This can be a source of bugs!
consider the following code:
```C++
for (auto ch : word) {  // Equivalent to: for (char& ch : word)
    ch += 127;
    count[ch]++;
}
```
This will change every character of the string, because `ch` is a reference to the character in the string! A bug that can be hard to chase down, especially in a large system.
Using `char` explicitly solves this.

annnufan:
Next Iteration

annnufan/assignment1/solution_1.cpp
```diff
@@ -18,12 +18,12 @@ bool word_anagram(std::string word_1, std::string word_2, bool case_sens) {
...
 	//String used signed char. So, it needs to add 127 for each char for counting sort 
-	for (auto ch : word_1) {
-		count[ch + 127]++;
+	for (char ch : word_1) {
+		count[(unsigned char)ch]++;
 	}
-	for (auto ch : word_2) {
-		if (count[ch + 127] > 0) {
-			count[ch + 127]--;
+	for (char ch : word_2) {
```
Brusi: You can use unsigned char in the loop to avoid repeating casts later:
```C++
for (unsigned char ch : word_2) { ...
```

PR #8, opened 2018-05-21
Assigment1

TijanaSonjic:
Solutions for tasks Q1 and Q2.

### DIFF
--- a/TijanaSonjic/assignment1/Q1.cpp
+++ b/TijanaSonjic/assignment1/Q1.cpp
@@ -1,0 +1,141 @@
+#include <iostream>
+#include <string>
+#include <sstream>
+#include <iterator>
+#include <unordered_map>
+#include <algorithm>
+
+using namespace std;
+
+bool isWord(string word) {
+	for (int i = 0; i < word.length(); i++)
+		if (!((word[i] >= 'a' && word[i] <= 'z') || (word[i] >= 'A' && word[i] <= 'Z')))
+			return false;
+	return true;
+}
+
+// Function that checks whether two words are anagram of each other.
+// Both arguments must be words. It has better time complexity than function areTwoSentencesAnagrams.
+// 
+bool isAnagram(string word1, string word2, bool caseSensitive) {
+	if (!isWord(word1) || !isWord(word2))
+		return false;
+
+	if (word1.length() != word2.length())
+		return false;
+
+	unordered_map<char, int> lettersDifCount;
+
+	for (int i = 0; i < word1.length(); i++) {
+		char letter = word1[i];
+
+		if (!caseSensitive && letter < 'a')
+			letter += 'a' - 'A';		
+			
+		auto it = lettersDifCount.find(letter);
+		if (it != lettersDifCount.end())
+			it->second++;
+		else
+			lettersDifCount.insert(lettersDifCount.end(), pair<char, int>(letter, 1));
+	}
+
+	for (int i = 0; i < word2.length(); i++) {
+		char letter = word2[i];
+		
+		if (!caseSensitive && letter < 'a')
+			letter += 'a' - 'A';
+
+		auto it = lettersDifCount.find(letter);
+		if (it != lettersDifCount.end() && it->second > 0)
+			it->second--;
+		else
+			return false;
+	}
+
+	// if we come this far this means that those words are anagrams
+	return true;
+}
+
+vector<string> splitWordsBySpace(string sentence) {
+	istringstream iss(sentence);
+	vector<string> words{ istream_iterator<string>{iss}, istream_iterator<string>{}};
+
+	return words;
+}
+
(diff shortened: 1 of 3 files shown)


TijanaSonjic/assignment1/Q2.h
```diff
@@ -0,0 +1,97 @@
+#pragma once
+
+
+template <class T>
+class List{
```
Brusi: Nice, everything is encapsulated in this class which is ready to be shipped :)

TijanaSonjic/assignment1/Q2.h
```diff
@@ -0,0 +1,97 @@
...
+	else {
+		tail->next = new Element<T>(info);
+		tail = tail->next;
+	}
+	count++;
+}
+
+template <class T>
+List<T>::Element<T>* List<T>::getKthToLastElement(int position) {
+	if (position < 0 || position >= count)
```
Brusi: Boundary checks :+1:

TijanaSonjic/assignment1/Q2.h
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	// Move both pointers
+	while (firstPointer) {
+		firstPointer = firstPointer->next;
+		secondPointer = secondPointer->next;
+	}
+
+	return secondPointer;
+}
```
Brusi: Nothing to add here, solution is looks complete, short and clear to understand :)

TijanaSonjic/assignment1/Q1.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+		if (it != wordsDifCount.end() && it->second>0)
+			it->second--;
+		else
+			return false;
+	}
+
+	// if we come so far this means that those words are anagrams of each other
+	return true;
+}
```
Brusi: I like the solution of generating 'keys' out of words which count as equivalent for anagram sentences :)

TijanaSonjic/assignment1/Q1.cpp
```diff
@@ -0,0 +1,141 @@
...
+	for (int i = 0; i < words2.size(); i++) {
+		string key = generateKey(words2[i], caseSensitive);
+
+		auto it = wordsDifCount.find(key);
+
+		if (it != wordsDifCount.end() && it->second>0)
+			it->second--;
+		else
+			return false;
+	}
```
Brusi: Extra credit:
This section really resembles what you did above (lines 35-53) when finding a word-anagram.
I would not call it a code duplication because it is still very different (one counts characters and one counts key strings) but still, can you think of a way to encapsulate both in one function/class?
(No need to change this code or write down an answer, just a thought experiment and a possibly interesting discussion).

TijanaSonjic/assignment1/Q1.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+	unordered_map<char, int> lettersDifCount;
+
+	for (int i = 0; i < word1.length(); i++) {
+		char letter = word1[i];
+
+		if (!caseSensitive && letter < 'a')
+			letter += 'a' - 'A';		
+			
+		auto it = lettersDifCount.find(letter);
```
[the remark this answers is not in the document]
TijanaSonjic: Great! :) Thank you!

TijanaSonjic/assignment1/Q1.cpp
```diff
@@ -0,0 +1,141 @@
...
+			return false;
+	}
+
+	// if we come this far this means that those words are anagrams
+	return true;
+}
+
+vector<string> splitWordsBySpace(string sentence) {
+	istringstream iss(sentence);
+	vector<string> words{ istream_iterator<string>{iss}, istream_iterator<string>{}};
```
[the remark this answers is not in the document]
TijanaSonjic: Yes, definitely

TijanaSonjic/assignment1/Q1.cpp
```diff
@@ -8,12 +8,22 @@
 using namespace std;
 
 bool isWord(string word) {
-	for (int i = 0; i < word.length(); i++)
-		if (!((word[i] >= 'a' && word[i] <= 'z') || (word[i] >= 'A' && word[i] <= 'Z')))
+	for each (char letter in word){
```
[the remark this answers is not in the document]
TijanaSonjic: Yes, thank you, I have missed this thing. The code compiles and passes the tests even with this syntax :D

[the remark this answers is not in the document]
Brusi: Interesting...
I also have a friend which writes C++ with 'and' and 'or' binary operands. Something like:
```C++
if (a > 1 and b > 3 or c > 5) ...
```
I don't know how or why it works... My first instinct is to use the common, c-style syntax, but I will be happy to hear more about this mystery if you hear anything :)

PR #9, opened 2018-05-29
assignment 2 solution

michbl:


michbl/assignment2/BinaryTree.hpp
```diff
@@ -0,0 +1,194 @@
+#ifndef BINARYTREE_HPP
+#define BINARYTREE_HPP
+
+#include <iostream>
+using namespace std;
+
+// Assuming T is comparable and printable
```
Brusi: It is good that you documented what are the preconditions of T.
This class also assumes that T is copyable :)

michbl/assignment2/BinaryTree.hpp
```diff
@@ -0,0 +1,194 @@
...
+    void printTree(Node<T> *node) {
+        if (node == nullptr) {
+            return;
+        }
+        printTree(node->left);
+        printTree(node->right);
+        cout << node->key << " ";
+    }
+
+    void printAncestors(Node<T> *node, T key, bool &found) {
```
Brusi: You do not change `node` in this function, correct?
So you can pass `node` as a const pointer.
`void printAncestors(const Node<T> *node, T key, bool &found)`
This checks in compile time that `node` is not changed; but also provides better documentation for the user of this function.

michbl/assignment2/BinaryTree.hpp
```diff
@@ -0,0 +1,194 @@
...
+        if (node == nullptr) {
+            return;
+        }
+        if (node->key == key) {
+            found = true;
+            return;
+        }
+        if (!found) {
```
Brusi: At this point `found` is always false, so this `if` is redundant.

-----
Additionally, you can use the return value instead of `found` output parameter. It can help the function be more concise:
```C++
bool found = printAncestors(node->left, key);
if (!found) {
  found = printAncestors(node->right, key);t
}
if (found) {
    cout << node->key << ", " << endl;  // I got lazy about printing the trailing comma :]
    return true;
}
return false;
```
Or even more concise:
```C++
if (printAncestors(node->left, key) || printAncestors(node->right, key)) {
    // The key is found in our left or right subtree.
    cout << node->key << ", " << endl;
    return true;
}
return false;
```
The latter is shorter and has the same logic; but has the downside that `printAncestors` have side effects (prints the value). Its function name contains `print` so it helps user not to overlook it.

michbl/assignment2/BinaryTree.hpp
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    bool bothKeysAreInSubtree(Node<T> *node, T key1, T key2) {
+        bool found1 = false, found2 = false;
+        bothKeysAreInSubtreeAux(node->left, key1, key2, found1, found2);
+        bothKeysAreInSubtreeAux(node->right, key1, key2, found1, found2);
+        return found1 && found2;
+    }
+
+    void bothKeysAreInSubtreeAux(Node<T> *node, T key1, T key2, bool &found1, bool &found2) {
```
Brusi: Tip:
You can use an additional helper function which checks if only one key is in a subtree, then `bothKeysAreInSubtree` function can call this function twice, once for each key.
I bet that it can make the code a little shorter (blind bet, I may be wrong).
Also, I think that a function for a single key is more valuable in terms of it can be used out of this context.

michbl/assignment2/BinaryTree.hpp
```diff
@@ -0,0 +1,194 @@
...
+            found1 = true;
+        }
+        if (node->key == key2) {
+            found2 = true;
+        }
+        bothKeysAreInSubtreeAux(node->left, key1, key2, found1, found2);
+        bothKeysAreInSubtreeAux(node->right, key1, key2, found1, found2);
+    }
+
+    void printLowestCommonAncestor(Node<T> *node, T key1, T key2, bool &found) {
```
Brusi: Same comment about using the return value.

michbl/assignment2/BinaryTree.hpp
```diff
@@ -0,0 +1,194 @@
...
+            cout << node->key << endl;
+        }
+    }
+
+
+public:
+    BinaryTree() {
+        this->root = nullptr;
+    }
+    BinaryTree(Node<T> *root) {
```
Brusi: At this moment, nodes are encapsulated in the tree and nothing returns them, yet this function gets a node as parameter. I don't think this constructor can ever be used at its current signature (unless constructing nodes outside of a tree, but I don't see a reason why to encourage the users to do so.

Is that a copy constructor?
If so, then its signature should be `BinaryTree(const Node<T>& root)`.
Anyway, this is optional because this assignment does not require you to copy trees :)

michbl:
Hi Ori,
I took your tips on making an Enum for 'side' and using functions' return values.
I also changed the function 'bothKeysAreInSubtree' to a function that takes only one key, to make it reusable.

About LCA, I saw someone asked on Slack about what happens if node1 is the ancestor of node2. The answer she got was to return node1. I disagree with that since I don't think a node is its own ancestor, so in that case, my implementation would return node1's parent. If I'm wrong I can easily fix it :)

michbl/assignment2/BinaryTree.hpp
```diff
@@ -84,111 +82,99 @@ class BinaryTree {
...
+        return (foundInLeft || foundInRight);
     }
 
 
 public:
     BinaryTree() {
         this->root = nullptr;
     }
-    BinaryTree(Node<T> *root) {
+    BinaryTree(const BinaryTree &tree) {
```
Brusi: Now `if (root == nullptr)` is going to check the local variable `root`.
Did you mean `tree`?
I how you did not because `tree` is a reference and can never be null :)
In that case, `this->root = copyTree(tree.root);` is enough for this function, because `copyTree` already handles the case in which `tree.root` is null, so no need to check this explicitly :)

michbl/assignment2/BinaryTree.hpp
```diff
@@ -84,111 +82,99 @@ class BinaryTree {
...
-            printLowestCommonAncestor(node->right, key1, key2, found);
-        }
-        if (!found && bothKeysAreInSubtree(node, key1, key2)) {
-            found = true;
-            cout << node->key << endl;
+            return false;
+        }
+        bool foundInLeft = printLowestCommonAncestor(node->left, key1, key2);
+        bool foundInRight = printLowestCommonAncestor(node->right, key1, key2);
+        if (!foundInLeft && !foundInRight) {
```
Brusi: The check `!foundInLeft && !foundInRight` is exactly the opposite of `foundInLeft || foundInRight` (according to De Morgans law). So this is a (minor) code duplication!
One option is to put it inside an else:
```C++
if (!foundInLeft && !foundInRight) {
    ...
} else {
    return true;
}
```
Or even remove the else because the first block always returns:
```C++
if (!foundInLeft && !foundInRight) {
    ...
}
return true;
```

But my suggestion is to flip the cases and return early. This way you will also have less nesting, and less need to navigate between all the ifs and elses:
```C++
if (foundInLeft || foundInRight) {
    // Both keys found in one of the subtrees, meaning this is an ancestor bot not
    // the lowest common ancestor.
    return true;
}
if (isAncestor(node, key1, key1)) {
    // This is the lowest common ancestor.
    cout << node->key;
    return true;
}
// This is not an ancestor at all.
return false;
```
This (optional) style avoids nesting and even without the comments, it is read as a list of cases and not a forest of if/elses (and no duplication). It has its downsides (e.g. many return points) but it is an option to consider.

michbl:
Hi Ori,

Made some fixes, I think the code really does look a lot better, especially "findLowestCommonAncestor" after I reversed the order of the if's.

I also separated the in/out code from the logic

Thanks!

PR #11, opened 2018-06-03
Assignment 2

TijanaSonjic:
Solutions for second assignment.


TijanaSonjic/assignment2/Q1.h
```diff
@@ -0,0 +1,74 @@
...
+	Node* getLeftChild() {
+		return leftChild;
+	}
+
+	Node* getRightChild() {
+		return rightChild;
+	}
+};
+
+// Function will return false if key is not in the tree.
```
Brusi: Tip:
Avoid double negative (in both code and comments):
`// Function will return true if key is in the tree.`
Is easier to understand.

TijanaSonjic/assignment2/Q2.h
```diff
@@ -0,0 +1,88 @@
+#pragma once
+#include <iostream>
+using namespace std;
+
+class Node {
```
Brusi: We encourage you to avoid duplication - so why submit the same code twice?
This `Node` class is the same in both files.
You can submit both questions in one file if that's more convenient for you (and use one `Node` class).
Another option is to extract Node into its own file and `#include` it in both solution files.

It will also make iterations easier for you (if you want to make a change in the node class, you do not need to change it twice).

TijanaSonjic/assignment2/Q2.h
```diff
@@ -0,0 +1,88 @@
...
+	if (!root)
+		return false;
+
+	if (root == node)
+		return true;
+
+	return isInTheTree(root->getLeftChild(), node) || isInTheTree(root->getRightChild(), node);
+}
+
+Node* findFirstCommonAncestor(Node* root, Node* node1, Node* node2) {
```
Brusi: If that is a helper function not intended to be called directly, there are several options to protect the users from using it:
1. Include it only in .cpp file and not in .h file (as an implementation detail)
2. If part of a class, declare it as private
3. Add a comment that clarifies this: `// This is a helper function, do not call directly. Call firstCommonAncestor instead.`
4. Give it a name that clarifies that this is a helper function, e.g. `firstCommonAncestorAux`.

I think in this context, 3 and 4 are good options. (No need to separate solutions to .h and .cpp in the assignments)

The point is to think of possible future users of this code, and think what we can do in order to make everyone use the code as we intended them to, making it hard to use it wrong. Calling this function directly may lead to unexpected results.

---

Also, this function assumes that both nodes are in the tree, so it is best to write this in a comment (to help the reader understand this function as an autonomous block).

TijanaSonjic/assignment2/Q1_test.cpp
```diff
@@ -0,0 +1,62 @@
...
+		cout << "Key is not in the tree." << endl;
+	}
+	cout << endl;
+
+	cout << "Ancestors of node with key value 55: ";
+	if (!printAncestors(nullRoot, 11)) {
+		cout << "Key is not in the tree." << endl;
+	}
+	cout << endl;
+
```
Brusi:
```C++
delete root;
```
You already wrote the destruction logic, so why not use it? :)

TijanaSonjic/assignment2/Q2.h
TijanaSonjic: Actually findFirstCommonAncestor is a helper function and as you said this function assumes that both nodes are in the tree, so it is called from firstCommonAncestor function. Function firstCommonAncestor will check if any of function input arguments are null and if they are not null it will check if both nodes are in the tree before we call findFirstCommonAncestor which is the function with the main logic.  
In the future I will write comments like this one you have mentioned in third option, they are good practice and make code more readable. :)

TijanaSonjic/assignment2/Q2.h
TijanaSonjic: I don't know :) I had a dilemma should I put those solution in separate files or can put all data structures in one file and just include it in both solutions. I was supposed to ask you this before submitting mu solutions.

TijanaSonjic/assignment2/Q1_test.cpp
TijanaSonjic: Oops :/

TijanaSonjic/assignment2/Q1.h
TijanaSonjic: I will do it in the future, thank you for drawing attention to that.

PR #12, opened 2018-06-03
Add assigment 2

annnufan:
First Iteration

### DIFF
--- a/annnufan/assignment2/node.cpp
+++ b/annnufan/assignment2/node.cpp
@@ -1,0 +1,96 @@
+#include <iostream>
+#include <algorithm>
+
+
+class node {
+	int key;
+public:
+	node(int x, node* l = nullptr, node* r = nullptr) : key(x), left_node(l), right_node(r){}
+	~node() {
+		delete left_node;
+		delete right_node;
+	}
+
+	void set_value(int x) {
+		key = x;
+	}
+
+	int get_value() {
+		return key;
+	}
+
+	node* left_node, *right_node, *parent_node;
+};
+
+bool find_key_and_print_way_to_root(int x, node* v) {
+	if (v == nullptr) {
+		return false;
+	}
+	if (x == v->get_value()) {
+		return true;
+	}
+	if (find_key_and_print_way_to_root(x, v->left_node) || find_key_and_print_way_to_root(x, v->right_node)) {
+		std::cout << v->get_value() << ' ';
+		return true;
+	}
+	return false;
+}
+
+void print_way_to_root_for_key(int x, node* v) {
+	if (find_key_and_print_way_to_root(x, v)) {
+		std::cout << "\n";
+	} else {
+		std::cout << "Couldn't find key\n";
+	}
+}
+
+node* find_node(int x, node* v, node* p) {
+	if (v == nullptr) {
+		return nullptr;
+	}
+	v->parent_node = p;
+	if (x == v->get_value()) {
+		return v;
+	}
+	node* u = find_node(x, v->left_node, v);
+	if (u != nullptr) {
+		return u;
+	}
+	u = find_node(x, v->right_node, v);
+	return u;
+}
+
+int count_high(node* v) {
+	if (v == nullptr) {
+		return -1;
+	}
+	int ans = 0;
+	while (v->parent_node != nullptr) {
+		ans++;
+		v = v->parent_node;
+	}
+	return ans;
+}
+
+int find_lca(int x, int y, node* root) {
+	node* u = find_node(x, root, nullptr);
+	node* v = find_node(y, root, nullptr);
+	if (u == nullptr || v == nullptr) {
+		return -1;
+	}
+	int u_h = count_high(u);
+	int v_h = count_high(v);
+	if (u_h < v_h) {
+		std::swap(u_h, v_h);
+		std::swap(u, v);
+	}
+	while (u_h > v_h) {
+		u_h--;
+		u = u->parent_node;
+	}
+	while (u != v) {
+		u = u->parent_node;
+		v = v->parent_node;
+	}
+	return u->get_value();
+}
--- a/annnufan/assignment2/node.cpp
+++ b/annnufan/assignment2/node.cpp
@@ -1,63 +1,74 @@
 #include <iostream>
 #include <algorithm>
 
-
 class node {
 	int key;
+	node *left_node, *right_node, *parent_node;
 public:
-	node(int x, node* l = nullptr, node* r = nullptr) : key(x), left_node(l), right_node(r){}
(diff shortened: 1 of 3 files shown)


annnufan/assignment2/node.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+void print_way_to_root_for_key(int x, node* v) {
+	if (find_key_and_print_way_to_root(x, v)) {
+		std::cout << "\n";
+	} else {
+		std::cout << "Couldn't find key\n";
+	}
+}
+
+node* find_node(int x, node* v, node* p) {
```
Brusi: This function not only finds nodes, but also fills the parent data.
From the perspective of a user, this can be unexpected; The user expects this function to return the node which has the key `x`. But this function also changes the state of the tree!
Also with the current function signature, it is not clear what the parameter `p` does.
A better name would be `find_node_and_update_parent`.

But even better:
Code is most readable and easy to reuse when each function does only one thing.
So you can split it to two functions:
`find_node` and `update_parents`, each one does just the one thing that you expect it to do from its name.

-----
But let's take a look at the `parent_node` field. When is it set?
Because it is sometimes set and sometimes not, it is hard for the user to follow when it is set and when they can use the functions that depend on it (e.g. `count_high`).
So I suggest that you update the `parent_node` field in the constructor; this way the user does not need to think - did I update this field? Did I forget? Why the code does not work? Because I forgot to update the parents! etc...

The constructor can work like this:
```C++
node(int x, node* l, node* r) : key(x), left(l), right(r) {
    if (left != null) left->parent = this;
    if (right != null) right->parent = this;
}

node* root = new node(1, new node(2, nullptr, nullptr), new node(3, nullptr, nullptr));
// After the line above, nodes 2 and 3 already have their parent field set correctly.
```
`parent` can stay private field because you only modify it inside `node` constructor.

annnufan/assignment2/node.cpp
```diff
@@ -0,0 +1,96 @@
...
+	}
+	node* u = find_node(x, v->left_node, v);
+	if (u != nullptr) {
+		return u;
+	}
+	u = find_node(x, v->right_node, v);
+	return u;
+}
+
+int count_high(node* v) {
```
Brusi: This function assumes that `parent_node` is set (which is not always the case at this moment).
So you can add a comment that clarifies this:
```C++
// Assumes that parent_node is set for this node and its parents.
int count_high(...
```

But better: see tip above to make sure that `parent_node` is always set, so this function always works as expected.

annnufan/assignment2/solution_1_test.cpp
```diff
@@ -0,0 +1,12 @@
...
+int main() {
+	node* root = new node(7, new node(3, new node(2, new node(1), new node(6)), new node(5)), new node(4, nullptr, new node(8)));
+	node* empty_root = nullptr;
+	print_way_to_root_for_key(7, root);
+	print_way_to_root_for_key(10, empty_root);
+	print_way_to_root_for_key(9, root);
+	print_way_to_root_for_key(1, root);
+	print_way_to_root_for_key(8, root);
+	return 0;
+}
```
Brusi: You can call the destructor at the end of this block to destroy the tree:
`delete root;`
Another option is to make `root` a local variable, so it is deleted implicitly when exiting the block:
`node root(7, new node(3, new node(2, ....`

annnufan/assignment2/node.cpp
Brusi: You already have a function above `find_key_and_print_way_to_root`. Maybe you can extract common logic to a function and avoid code duplication? For example, both solutions can re-use the same `find_key` function.

annnufan/assignment2/node.cpp
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	int get_value() {
+		return key;
+	}
+
+	node* left_node, *right_node, *parent_node;
+};
+
+bool find_key_and_print_way_to_root(int x, node* v) {
```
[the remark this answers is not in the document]
annnufan: I aree about key. But I am not sure about `tree`. As I know, `v` and `u` is common name for  nodes in tree.
Or is  it correct only for quick solves studing problems?

annnufan:
As I see, I corrected code as writing in most comments.
Now I'm ready for next iteration.

annnufan/assignment2/node.cpp
```diff
@@ -1,80 +1,91 @@
...
 int count_high(node* v) {
 	if (v == nullptr) {
 		return -1;
 	}
 	int ans = 0;
-	while (v->parent_node != nullptr) {
+	while (v->get_parent() != nullptr) {
 		ans++;
-		v = v->parent_node;
+		v = v->get_parent();
```
Brusi: Tip: You can get the same result like this:
```C++
int count_high(node* v) {
    int ans = -1;
    while (v != null) {
        ans++;
        v = v->get_parent();
    }
    return ans;
}
```
Slightly shorter, but more importantly, only one call to `get_parent`.

annnufan/assignment2/node.cpp
```diff
@@ -1,80 +1,91 @@
...
 	}
 	return ans;
 }
 
-int find_lca(int x, int y, node* root) {
-	node* u = find_node(x, root, nullptr);
-	node* v = find_node(y, root, nullptr);
+int find_lowest_common_ancestor(int key_1, int key_2, node* root) {
+	node* u = find_node(key_1, root);
```
Brusi: More readable now that there is no extra `nullptr` parameter 👍

annnufan/assignment2/solution_2_test.cpp
```diff
@@ -2,12 +2,11 @@
...
-	node* root = new node(7, new node(3, new node(2, new node(1), new node(6)), new node(5)), new node(4, nullptr, new node(8)));
-	node* empty_root = nullptr;
-	assert(find_lca(7, 8, root) == 7);
-	assert(find_lca(10, 9, empty_root) == -1);
-	assert(find_lca(6, 9, root) == -1);
-	assert(find_lca(1, 6, root) == 2);
-	assert(find_lca(8, 6, root) == 7);
+	node root(7, new node(3, new node(2, new node(1), new node(6)), new node(5)), new node(4, nullptr, new node(8)));
+	assert(find_lowest_common_ancestor(7, 8, &root) == 7);
+	assert(find_lowest_common_ancestor(10, 9, nullptr) == -1);
```
Brusi: In my opinion the line:
`node* empty_root = nullptr;`
helped readability because this made this line clear.
(`find_lowest_common_ancestor(10,9,empty_tree);` make the intention of this test very clear).

But `nullptr` is also ok (the reader at this point should know that `nullptr` represents an empty tree)

PR #15, opened 2018-06-12
assignment 3 solution

michbl:


michbl/assignment3/main.cpp
```diff
@@ -0,0 +1,43 @@
...
+#include "Grid.hpp"
+#include "codeu_test_lib.h"
+
+int main() {
+    vector<char> g = {'A', 'A', 'R', 'T', 'C', 'D'};
+    Grid grid(g, 2, 3);
+
+    // test assignment example
+    vector<string> words = {"CAR", "CARD", "CART", "CAT"};
+    Dictionary dict1(words);
```
Brusi: Tip:
You can inline the words list:
```C++
Dictionary dict1({"CAR", "CARD", "CART", "CAT"});
```
It is really up to your personal preference :)

michbl/assignment3/Dictionary.hpp
```diff
@@ -0,0 +1,128 @@
...
+    void wordSearchBT(Grid &grid, int row, int col, set<string> &res, string word) {
+        char c;
+        if (!grid.getChar(row, col, c)) {
+            return;
+        }
+        word += c;
+        if (isWord(word)) {
+            res.insert(word);
+        }
+        if (isPrefix(word)) {
```
Brusi: Tip:
You can do:
```
if (!isPrefix(word)) {
    return;
}
```
To avoid nesting the rest of the function.

michbl/assignment3/Dictionary.hpp
```diff
@@ -0,0 +1,128 @@
...
+
+    Node *root;
+    vector<Node *> nodes; // for memory release
+
+    void wordSearchBT(Grid &grid, int row, int col, set<string> &res, string word) {
+        char c;
+        if (!grid.getChar(row, col, c)) {
+            return;
+        }
+        word += c;
```
Brusi: Tip:
You can pass `word` as a reference (or pointer) to this function, to avoid copying it again and again every function call (can be very inefficient if this code runs for very long words).
Upside: more efficient!
Downside: This way you need to remember to remove every letter after you added it, before returning from the function. (less rubust and less readable).

```C++
word.push_back(c);  // Adds c on the same buffer without allocating a new one.
if (isPrefix()) {
   for (Direction dir : dirs) { /* recursive call*/ }
}
word.pop();
return;
```

michbl/assignment3/Dictionary.hpp
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    ~Dictionary() {
+        for (auto node : this->nodes) {
+            delete node;
+        }
+    }
+
+    void insertWord(const string &word) {
+        unordered_map<char, Node *> *currentChildren = &(this->root->children);
```
Brusi: Tip:
As far as I know, C++ operator precedence rules let you remove the parenthesis:
`&this->root->children;`
here and below.

michbl/assignment3/Dictionary.hpp
```diff
@@ -0,0 +1,128 @@
...
+        unordered_map<char, Node *>::iterator it;
```
Brusi: The benefit of defining a variable before the `for` loop is efficiency - instead of creating the same variable over and over, it is created once and then overridden.
The downside is readability - it is harder to understand the context of this iterator. The following is more readable in my opinion:
```C++
for (char ch : sord) {
    unordered_map<char, Node *>::iterator it = currentChildren->find(ch);
    // Can also be:    auto it = currentChildren->find(ch);
    if (it != ...
   ...
}
```
but can be less efficient, especially if the repeating variable is heavy to construct or allocate.

Both are OK, and it is good to consider efficiency when writing code. It is up to you to decide which fits in each case. Just remember:
> *"Premature optimization is the root of all evil" - Donald Knuth*

;)

michbl/assignment3/Dictionary.hpp
```diff
@@ -0,0 +1,128 @@
...
+#include <string>
+#include <unordered_map>
+#include <set>
+#include "Grid.hpp"
+
+#define END_OF_WORD -1
+
+using namespace std;
+
+// implementing dictionary as trie
```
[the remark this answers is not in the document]
michbl: I did understand that we're not required to implement a dictionary, but I didn't know how to test my solution without implementing one. Do you mean there is already a dictionary with these functions in some library, or do you mean I could have implemented it non-efficiently?
Anyways, isPrefix made me think immediately about trie so I thought I'd try to write it, and besides, it's in the optional follow-up :)
Sorry I gave you extra work 😳

michbl/assignment3/Dictionary.hpp
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    bool isWord(const string &word) {
+        unordered_map<char, Node *> lastNodeChildren;
+        lastNodeChildren.clear();
+        bool prefix = isPrefix(word, &lastNodeChildren);
+        return prefix && lastNodeChildren.find(END_OF_WORD) != lastNodeChildren.end();
+    }
+
+    set<string> wordSearch(Grid grid) {
```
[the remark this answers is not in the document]
michbl: I'm convinced, I'll separate the function from the class

michbl/assignment3/Dictionary.hpp
michbl: The downside is why I wrote it like that. Do you think the upside wins in this case?

michbl/assignment3/Dictionary.hpp
michbl: I'll keep that in mind from now on :)

michbl/assignment3/Dictionary.hpp
```diff
@@ -0,0 +1,128 @@
...
+        char c;
+        unordered_map<char, Node *> children;
+
+        Node(char c) : c(c) {
+            this->children.clear();
+        }
+    };
+
+    Node *root;
+    vector<Node *> nodes; // for memory release
```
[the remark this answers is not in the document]
michbl: I deleted that vector that held all the nodes and added destroyDict() which deletes all nodes recursively

michbl/assignment3/Dictionary.hpp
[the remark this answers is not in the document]
Brusi: Don't worry about the extra work ^^
I'm reviewing happily, just wanted to make sure that you know which part of the assignment is mandatory and which is optional. (You are more than welcome to write additional code to the assignment, like writing your own data structures or testing framework or anything that was not a direct requirement of the assignment, it means more to review and discuss which is good - but it really is up to you and not mandatory).

As for the dictionary, you could use practically anything that has `isPrefix` and `isWord` functions that act as expected, with complete freedom on its implementation or initialization.
For example, the dictionary could be simply two sets, one of prefixes and one of words, which you insert manually in the tests:
```C++
class Dictionary {
    set<string> words;
    set<string> prefixes;
  public:
    Dictionary(const set<string>& words, const set<string>& prefixes) 
        : words(words), prefixes(prefixes) {}
    
    bool isPrefix(const string& s) { return prefixes.find(s) != prefixes.end(); }
    bool isWord(const string& s) { return words.find(s) != words.end(); }
};

// Dictionary initialization (in the tests):
Dictionary dict({"CAT","CAR", "CARD"},
                {"C","CA","CAT","CAR","CARD"});
```
This is the simplest I can think of. Any other dictionary implementation with same post-initialization interface would work.

Your solution can assume that the dictionary is legal (that `isPrefix(s)` returns true iff `s` is a prefix of a `t` such that `isWord(t)` returns true). Dictionary implementation does not need to enforce this (in this assignment scope).

michbl/assignment3/Dictionary.hpp
Brusi: Hmm, it really depends.
If we follow "Premature optimization is the root of all evil", then readability should be our first instinct. I am personally in favor of "functional" programming, making each function independent and share the least info between them.
In this case, this is a recursive call which can happen many times, and copying the string can have a real effect of the program runtime; but even this depends on the input (how long the words are, how large is the board...).

I think that most important is that you understand the implications of both ways.
If an interviewer will ask you what is the complexity of this solution, then you will be able to spot the cost of copying the word; and when they ask "can you do better", they you will say: "Yes, we can share one accumulating word between the functions".

michbl/assignment3/main.cpp
```diff
@@ -1,42 +1,128 @@
...
+enum Direction {
+    LEFT, RIGHT, UP, DOWN, TOP_LEFT, TOP_RIGHT, BOTTOM_LEFT, BOTTOM_RIGHT
+};
+
+// function declarations
+void getNextIdxs(const int row, const int col, const Direction dir, int &nextRow, int &nextCol);
+
+set<string> wordSearch(const Dictionary &dict, const Grid &grid);
+
+void wordSearchBT(const Dictionary &dict, const Grid &grid, int row, int col, set<string> &res, string word);
```
Brusi: No need to separate declarations and definitions for these assignments. You can mostly avoid the need for declarations by writing the functions in the order you use them.
(Unless required by the compiler for some reason, e.g. function A which calls functions B which calls function A)

michbl:
Oops, I completely forgot about that requirement!
Added a boolean vector to Grid to keep track of visited cells

PR #16, opened 2018-06-17
First iteration for 3 assignment

annnufan:
First iteration for solving problem with grid


annnufan/assignment3/founder.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+class founder {
+	dictionary dict;
+	std::vector<std::vector<char>> grid;
+	std::set<std::string> words_from_grid;
+	bool find_before;
+	int n, m;
+	std::vector<std::vector<int>> moves;
+
+	void go_from_cell(int i, int j, std::vector<std::vector<bool>> used, std::string word) {
```
Brusi: Here you pass `used` and `word` by value, which means that they are copied for each recursive call. It can greatly increase the complexity of the solution, especially if the grid is large.

An alternative which is more efficient is to have one copy for each of `used` and `word`, and pass it by pointer/reference to each recursive call.
If you choose to do so, then you need to remember to update them before and after each recursive call:
```C++
// At the beginning:
word.push_back(grid[i][j]);
used[i][j] = true;
......
// Before returning:
word.pop_back();
used[i][j] = false;
```

This solution is more efficient (no copies of large vectors for each function call), but can be less robust - forgetting to set `word` and `used` back can cause bugs. So be aware of both when choosing your solution.

annnufan/assignment3/founder.cpp
```diff
@@ -0,0 +1,74 @@
...
+		word += grid[i][j];
+		used[i][j] = true;
+		if (!dict.is_prefix(word)) {
+			return;
+		}
+		if (dict.is_word(word)) {
+			words_from_grid.insert(word);
+		}
+		for (auto k : moves) {
```
Brusi: This version of `for` copies each move into `k`. Iterating on a reference can solve that. Also, if you do not want to change the `moves` array, you can declare `k` as const reference for safety:
```C++
for (const auto& k : moves) { ...
```

Also... Why the name `k`? `i` and `j` are indices, `n` and `m` are sizes... But what is `k`?
I think a more descriptive name can be used, e.g. `move`.
`for (const auto& move : moves)` is more readable in my opinion.

annnufan/assignment3/test_founder.cpp
```diff
@@ -0,0 +1,31 @@
...
+	EXPECT_EQ(find_words_from_dict_on_grid(ex_dict, ex_grid), correct_ans);
+}
+
+void test_empty_dict() {
+	EXPECT_EQ(find_words_from_dict_on_grid(empty_dict, ex_grid), empty_ans);
+}
+
+void test_empty_grid() {
+	EXPECT_EQ(find_words_from_dict_on_grid(ex_dict, empty_grid), empty_ans);
+}
```
Brusi: Tip (optional):
You can add a test that makes sure that each letter can be used only once per word.

annnufan/assignment3/founder.cpp
annnufan: Yeah, I thought about both options and choose more readable option for this case withount "hard" clear and update after usage.

annnufan:
I make some changes :)

annnufan/assignment3/test_words_finder.cpp
```diff
@@ -0,0 +1,39 @@
...
+#include "codeu_test_lib.h"
+#include "words_finder.cpp"
+
+std::vector<std::vector<char>> ex_grid({{'a', 'a', 'r'}, {'t', 'c', 'd'}});
+std::vector<std::vector<char>> empty_grid;
+std::vector<std::vector<char>> grid_1({{'a', 'o', 'b', 'e', 'r'}, {'c', 'a', 'd', 'e', 't'}, {'b', 'i', 'v', 'i', 'n'}});
+
+dictionary ex_dict({"car", "card", "cart", "cat"});
+dictionary empty_dict;
+dictionary dict_1({"tree", "cart", "bee", "bed", "cider", "beer", "teer", "dinner", "baobab", "vine", "tear", "horse", "bob"});
```
Brusi: If not shared between tests, it is recommended to define test input and expected output inside the test case. This way the reader can see the entire test in its context without the need to scroll up and down:
```C++
void test_double_usage_cell() {
    std::vector<std::vector<char>> grid = {{'s', 'o', 'm', 'e'}
                                           {'g', 'r', 'i', 'd'}};
    dictionary dict({"some", "words"});
    std::set<std::string> expected_result = {...};
    EXPECT_EQ(expected_result, find_words_from_dict_on_grid(dict, grid));
}
```

Even if the input IS shared between tests, some say it is ok to copy-paste it to all the tests so each test will be easily readable by itself. (Yes, code duplication is more forgivable in tests!)

annnufan:
Third iteration

annnufan/assignment3/test_words_finder.cpp
```diff
@@ -2,32 +2,35 @@
...
+	std::set<std::string> empty_ans;
+
 	EXPECT_EQ(find_words_from_dict_on_grid(ex_dict, empty_grid), empty_ans);
 }
 
 void test_double_usage_cell() {
+	dictionary dict_1({"tree", "cart", "bee", "bed", "cider", "beer", "teer", "dinner", "baobab", "vine", "tear", "horse", "bob"});
+	std::vector<std::vector<char>> grid_1({{'a', 'o', 'b', 'e', 'r'}, {'c', 'a', 'd', 'e', 't'}, {'b', 'i', 'v', 'i', 'n'}});
+	std::set<std::string> ans_1({"tree", "bee", "bed", "cider", "beer", "teer", "vine"});
+
```
Brusi: I think that this way, each test is much more readable, as you can easily see its input and expected output.
Well done! :+1: 

-----
FYI
In other cases, where the input initialization is very long and tests end up having a lot of common "copy-pasted" code, it is possible to extract the common part to a function. That's not the case here, because code defining the input is quite short (2 lines), and having the code a bit copied is not too bad.
For example, lines 6 and 22 are the same, but still I prefer it this way, not extracted to some common function or variable, so each test can be independently read. Do you agree?

annnufan: Yes, I agree, that this type of declaration in this cases easier to understand

PR #18, opened 2018-06-18
Can't visit same cell twice

michbl:
Added tracking of visited cells and a test for that requirement

I also think I made a little mess with the pull requests.. Anyways, this is the one that needs to be reviewed

### DIFF
--- a/michbl/assignment3/Grid.hpp
+++ b/michbl/assignment3/Grid.hpp
@@ -8,6 +8,7 @@ using namespace std;
 class Grid {
 private:
     vector<char> grid;
+    vector<bool> visited;
     int height;
     int width;
 public:
@@ -18,6 +19,7 @@ public:
         this->grid = g;
         this->height = h;
         this->width = w;
+        this->clearVisited();
     }
 
     ~Grid() {}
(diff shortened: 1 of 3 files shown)


michbl/assignment3/Grid.hpp
```diff
@@ -39,6 +41,37 @@ class Grid {
         c = this->grid[idx];
         return true;
     }
+
+    void clearVisited() {
```
Brusi: Every function is best doing only one thing; same about classes!
The "visited"/"not visited" is a trait of the word search algorithm, not of the grid itself.
The grid class, when kept simple, can be reused: Between several word searches, for even used for other word games!
So I recommend decouple the visited status from the grid class. and make the visited status a local variable of the word search function.
You can still use a dedicated class for it! Called a VisitedGrid or a BinaryGrid, it can do everything this grid does (regarding the visited state) but not on the actual grid :)

Saving status on the grid can also potentially cause bugs, if one word search had changed the grid's state and forgot to restore the state, it can make word searches affect each other, and cause unexpected behavior.
But mostly it is for code reusability and readability.

PR #19, opened 2018-06-18
Solution for assignment 3

TijanaSonjic:


TijanaSonjic/assignment3/wordSearch.cpp
```diff
@@ -0,0 +1,160 @@
...
+	if (!isPrefix(prefixes, word)) {
+		return;
+	}
+
+	if (isWord(dictionary, word)) {
+		matches.push_back(word);
+	}
+
+	grid[i][j].visited = true;
+	vector<char> newWord(generatedWord);
```
Brusi: Why are you creating a new word here? `generatedWord` already has the value you want (the previous word plus the letter of the current grid). You can pass it directly to the recursive call of `findWords`.

TijanaSonjic/assignment3/wordSearch.cpp
```diff
@@ -0,0 +1,160 @@
...
+
+	for (int k = 0; k < 8; k++) {
```
Brusi: 8 is a "magic number"....
You can avoid it by making "difference" a vector, then you can iterate on its size:
`for (int k=0; k < difference.size(); k++) {`
Or better, use a foreach loop:
`for (const Delta& delta : difference) {
    bool valid_move = check(i + delta.dx, j + delta.dj);
    ....
}

TijanaSonjic/assignment3/wordSearch.cpp
```diff
@@ -0,0 +1,160 @@
...
+		bool validToMove = check(i + difference[k].dx, j + difference[k].dy);
```
Brusi: You use both "i,j", and "x,y" coordinates to represent location on the grid. This can be confusing, it is best to stick to one method and keep it.
It is even more confusing, because `i` represents the row, meaning changing of `i` is going up and down; and `j` represents the column, meaning changing `j` is going left and right;
So it means that `j` should correspond with `x`, and `i` should correspond with `y`!
Confused? Me too! ;)

Therefore I recommend changing `Delta` struct fields to be `d_i, d_j` or `diff_i,diff_j`, and not use `x,y` at all.

TijanaSonjic/assignment3/wordSearch.cpp
```diff
@@ -0,0 +1,160 @@
...
+	vector<string> empty;
+
+	for (int i = 0; i < result.size(); i++)
+		cout << result[i] << endl;
+
+	EXPECT_EQ(findAllWordsFormDictionaryInTheGrid(dictionary, grid), correctAnswers);
+
+	EXPECT_EQ(findAllWordsFormDictionaryInTheGrid(emptyDictionary, grid), empty);
+
+	EXPECT_EQ(findAllWordsFormDictionaryInTheGrid(dictionary, new GridCell[3][3]), empty);
```
Brusi: You can assume that all grid cells contain legal alphabetical values, so you don't have to support this case if a grid without content (I think this grid would be filled with zeros, but I am not sure).
If you want to check an empty grid, try to check if your code works on a grid of size 0x0.

TijanaSonjic/assignment3/wordSearch.cpp
```diff
@@ -0,0 +1,160 @@
...
+#include "vector"
+#include <algorithm>
+#include <string>
+#include "test_libraryr.h"
+
+using namespace std;
+
+
+
+set<string> dictionary = { "apple", "spring", "sun", "sea", "dancer" };
```
Brusi: Try to move the dictionary and the grid to be local variables of the tests, and pass them to your solution function, rather than global variables. This way the compiler would make sure that the solution does not access these globals, and would work for any dictionary/grid given to it.

TijanaSonjic/assignment3/wordSearch.cpp
```diff
@@ -0,0 +1,160 @@
...
+			findWords(dictionary, prefixes, matches, i + difference[k].dx, j + difference[k].dy, newWord);
+	}
+
+	grid[i][j].visited = false;
+	generatedWord.pop_back();
+}
+
+// If word can be formed in the grid, it will be included in resulting vector.
+//
+vector<string> findAllWordsFormDictionaryInTheGrid(set<string> dictionary, GridCell grid[][3]) {
```
Brusi: Your code gets `dictionary` as a set of strings, meaning that anyone who uses your code will have to implement their dictionary as a set of strings. What if you want to make it easy for the user to change the implementation of dictionary without having to modify this code?
Exactly for that, Object Oriented Programming comes in!
You can create a dictionary class, and *encapsulate* the implementation in the class, providing a simple interface:
```C++
class Dictionary {
  private:
    // Some fields, for example:
    std::set<string> words;
  public:
    // constructor
    Dictionary(const std::set<string>& words) ...

    // Interface functions, the any user of this function will use these functions only!
    bool isPrefix(const string& str) { ... }
    bool isWord(const string& str) { ... }
};
```

Then in your code, you can get the dictionary as a parameter:
vector<string> findAllWords(const Dictionary& dictionary, ...) {
   ...
   if (dictionary.isPrefix(word)) ...  // Example usage.
}

This case, your solution only uses `isPrefix` and `isWord` functions, and does not rely on the implementation of Dictionary. Now the next programmer which comes to change or use this code (which can be you in the future or someone else) can change the dictionary implementation (for exmple, if someone wants to improve its complexity), without the need to change anything of the "find words in grid" algorithm. They only have to make sure that at the end, `isPrefix` and `isWord` behave the same.
Easy! Portable! Encapsulated! Object Oriented! :-0

TijanaSonjic:
Code refactoring and migration into classes.

TijanaSonjic/assignment3/wordSearch.cpp
```diff
@@ -0,0 +1,158 @@
...
+#include <string>
+#include "test_libraryr.h"
+
+using namespace std;
+
+
+
+set<string> dictionary = { "apple", "spring", "sun", "sea", "dancer" };
+
+typedef struct gridCell {
```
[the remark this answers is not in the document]
TijanaSonjic: Thanks for pointing this out. I will keep in mind for future.

TijanaSonjic/assignment3/wordSearch.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+// Directions for adjacent cells
+Delta difference[8] = { {-1,-1},{-1,0},{-1,1},{0,-1},{0,1},{1,-1},{1,0},{1,1} };
+
+// Discard words that have letters which are not in the grid.
+//
+set<string> reduceDictionary(set<string> dictionary, set<char> gridLetters) {
+	set<string> result;
+
+	for (set<string>::iterator it = dictionary.begin(); it != dictionary.end(); ++it) {
```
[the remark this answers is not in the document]
TijanaSonjic: Although I know that, I see that for me it is so hard to change this habit of writing classic for loop.

TijanaSonjic/assignment3/wordSearch.cpp
```diff
@@ -0,0 +1,158 @@
...
+bool isPrefix(set<string> prefixes, string word) {
+	return prefixes.find(word) != prefixes.end();
+}
+
+bool isWord(set<string> dictionary, string word) {
+	return dictionary.find(word) != dictionary.end();
+}
+
+bool check(int i, int j) {
+	return i >= 0 && i < 3 && j >= 0 && j < 3 && !grid[i][j].visited;
```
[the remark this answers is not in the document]
TijanaSonjic: I totally agree with this.

TijanaSonjic/assignment3/wordSearch.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+	grid[i][j].visited = false;
+	generatedWord.pop_back();
+}
+
+// If word can be formed in the grid, it will be included in resulting vector.
+//
+vector<string> findAllWordsFormDictionaryInTheGrid(set<string> dictionary, GridCell grid[][3]) {
+	set<char> gridLetters = makeGridSet(grid);
+	set<string> reducedDictionary = reduceDictionary(dictionary, gridLetters);
```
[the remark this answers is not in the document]
TijanaSonjic: I made the assumption that I was not supposed to make. This is a great advice for an interview, thanks!

PR #21, opened 2018-06-19
Ayelet k 3

ayeletK:
Please review my code.

I chose to implement the grid for char (instead of a generic one).
My algorithm is a variant of backtracking.
I want to hear about alternative solutions and about better practice of my code.

Thanks!

### DIFF
--- a/ayeletK/assignment3/dictionary.cpp
+++ b/ayeletK/assignment3/dictionary.cpp
@@ -1,0 +1,28 @@
+#include <iostream>
+#include <string>
+#include <set>
+#include <vector>
+
+using namespace std;
+
+class Dictionary{
+    set<string> _words;
+    set<string> _prefixes;
+public:
+    Dictionary(set<string> words): _words(words) {
+        vector<char> curr_prefix;
+        for (string word: _words){
+            for (char c: word){
+                curr_prefix.push_back(c);
+                _prefixes.insert(string(curr_prefix.begin(), curr_prefix.end()));
+            }
+            curr_prefix.clear();
+        }
+    }
+    bool is_word(string word){
+        return _words.find(word) != _words.end();
+    }
+    bool is_prefix(string prefix){
+        return _prefixes.find(prefix) != _prefixes.end();
+    }
+};
--- a/ayeletK/assignment3/grid.cpp
+++ b/ayeletK/assignment3/grid.cpp
@@ -1,0 +1,120 @@
+#include "position.cpp"
+#include "dictionary.cpp"
+
+class Grid{
+    int _rows, _cols;
+    vector<vector<char>> _grid;
+    bool update_location(int* row, int* col){
+        if (*col + 1 < _cols){ 
+            (*col)++;
+            return true;
+        }
+        else if (*row + 1 < _rows) {
+            *col = 0;
+            (*row)++;
+            return true;
+        }
+        return false;
+    }
+    bool legal_row(int row){
+        return row >=0 && row < _rows;
+    }
+    bool legal_col(int col){
+        return col >=0 && col < _cols;
+    }
+    bool legal_position(Position pos){
+        return legal_row(pos._row) && legal_col(pos._col);
+    }
+    set<Position> get_neighbours(Position curr_pos){
+        Position neighbour;
+        set<Position> neighbours;
+        for(int row=curr_pos._row - 1; row <= curr_pos._row + 1; ++row){
+            for(int col=curr_pos._col - 1; col <= curr_pos._col + 1; ++col){
+                neighbour = Position(row, col);
+                if (legal_position(neighbour) && !curr_pos.equal(neighbour)) {
+                    neighbours.insert(neighbour);
+                }
+            }
+        }
+        return neighbours;
+    }
+    set<Position> get_unvisited_neighbours(Position curr_pos, set<Position>* visited){
+        set<Position> res;
+        for (Position pos: get_neighbours(curr_pos)){
+            if (visited->find(pos) == visited->end())
+                res.insert(pos);
+        }
+        return res;
+    }
+    void words_with_prefix(Position curr_pos, set<Position>* visited, 
+    Dictionary* dict, vector<char>* prefix, set<string>* words){
+        prefix->push_back(get_value(curr_pos));
+        visited->insert(curr_pos);
+        string curr_prefix = string(prefix->begin(), prefix->end());
+        if (! dict->is_prefix(curr_prefix)) {
+            visited->erase(curr_pos);
+            prefix->pop_back();
+            return;
+        }
+        if (dict->is_word(curr_prefix))
+            words->insert(curr_prefix);
+        for (Position neighbour: get_unvisited_neighbours(curr_pos, visited))
+            words_with_prefix(neighbour, visited, dict, prefix, words);
+        visited->erase(curr_pos);
+        prefix->pop_back();
+    }
+    set<string> find_words_from_position(Position start_pos, Dictionary* dict){
+        set<string> res;
+        vector<char> prefix;
+        set<Position> visited;  // can be improved to hashset ?
+        words_with_prefix(start_pos, &visited, dict, &prefix, &res);
+        return res;
+
+    }
+    set<string> find_all_words(Dictionary* dict){
+        set<string> res;
+        set<string> iteration_res;
+        for (int row = 0; row<_rows; ++row){
+            for (int col = 0; col < _cols; ++col){
+                iteration_res = find_words_from_position(Position(row,col),dict);
+                res.insert(iteration_res.begin(), iteration_res.end());
+            }
+        }
+        return res;
+    }
+public:
+    explicit Grid(int rows, int cols): _rows(rows), _cols(cols) {
+        // _grid.reserve(_rows);
+    }
+    void initialize_grid(vector<char> data){
+        int row = 0, col = 0;
+        for (int i = 0; i < data.size(); ++i){
+            if (col == 0){
+                _grid.push_back(vector<char>());
+            }
+            _grid.at(row).push_back(data.at(i));
+            if (!update_location(&row, &col)) break;
+        }
+    }
+    void update_value(int row, int col, char val){
+        if (legal_row(row) && legal_col(col))
+            _grid.at(row).at(col) = val;
+    }
+    char get_value(int row, int col){
+        if (legal_row(row) && legal_col(col))
+            return _grid.at(row).at(col);
+    }
+    char get_value(Position pos){
+        if (legal_position(pos))
+            return _grid.at(pos._row).at(pos._col);
+    }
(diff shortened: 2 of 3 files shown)


ayeletK/assignment3/grid.cpp
```diff
@@ -0,0 +1,120 @@
...
+            words->insert(curr_prefix);
+        for (Position neighbour: get_unvisited_neighbours(curr_pos, visited))
+            words_with_prefix(neighbour, visited, dict, prefix, words);
+        visited->erase(curr_pos);
+        prefix->pop_back();
+    }
+    set<string> find_words_from_position(Position start_pos, Dictionary* dict){
+        set<string> res;
+        vector<char> prefix;
+        set<Position> visited;  // can be improved to hashset ?
```
Brusi: In case the order of the elements in the set is not important, then a hash set (e.g. `std::unordered_set`) can be more efficient than a tree set (e.g. `std::set`).
However the difference between the sets it is not that important for me in the scope of this course, as they are both pretty efficient. As long as you make the correct logical use of sets, both C++ library sets are OK (and easily interchangeable too).

ayeletK/assignment3/grid.cpp
```diff
@@ -0,0 +1,120 @@
...
+        }
+        return res;
+    }
+    void words_with_prefix(Position curr_pos, set<Position>* visited, 
+    Dictionary* dict, vector<char>* prefix, set<string>* words){
+        prefix->push_back(get_value(curr_pos));
+        visited->insert(curr_pos);
+        string curr_prefix = string(prefix->begin(), prefix->end());
+        if (! dict->is_prefix(curr_prefix)) {
+            visited->erase(curr_pos);
```
Brusi: Tip:
You can insert the position to the `visited` set after the `is_prefix` check , so you will not need to erase it here.
Sadly it is not possible to do the same with the `prefix` string ;)

ayeletK/assignment3/grid.cpp
```diff
@@ -0,0 +1,120 @@
...
+        if (dict->is_word(curr_prefix))
+            words->insert(curr_prefix);
+        for (Position neighbour: get_unvisited_neighbours(curr_pos, visited))
+            words_with_prefix(neighbour, visited, dict, prefix, words);
+        visited->erase(curr_pos);
+        prefix->pop_back();
+    }
+    set<string> find_words_from_position(Position start_pos, Dictionary* dict){
+        set<string> res;
+        vector<char> prefix;
```
Brusi: Like above, `prefix` can be a string. You can still use the same `push_back`,`pop_back` interface.

ayeletK/assignment3/grid.cpp
```diff
@@ -0,0 +1,120 @@
...
+        visited->insert(curr_pos);
+        string curr_prefix = string(prefix->begin(), prefix->end());
+        if (! dict->is_prefix(curr_prefix)) {
+            visited->erase(curr_pos);
+            prefix->pop_back();
+            return;
+        }
+        if (dict->is_word(curr_prefix))
+            words->insert(curr_prefix);
+        for (Position neighbour: get_unvisited_neighbours(curr_pos, visited))
```
Brusi: I like how you encapsulated the iteration into a function which gets exactly what we want to iterate on. Makes this loop very readable :+1:

ayeletK/assignment3/grid.cpp
```diff
@@ -0,0 +1,120 @@
...
+                res.insert(iteration_res.begin(), iteration_res.end());
+            }
+        }
+        return res;
+    }
+public:
+    explicit Grid(int rows, int cols): _rows(rows), _cols(cols) {
+        // _grid.reserve(_rows);
+    }
+    void initialize_grid(vector<char> data){
```
Brusi: Suggestion for an alternative implementation: 
Instead of iterating on the data, iterate on the location:
```C++
void initialize_grid(vector<char> data){
    int data_pos = 0;
    for (int row=0; row < _rows; ++row) {
        _grid.push_back(vector<char>(cols));
        for (int col=0; col < _cols; ++col) {
            _grid[row].push_back(data[data_pos]);
            ++data_pos;
        }
    }
}
```
This spares you separating the row/col advancement logic into "update_location" function.

Another way is to just get the vector of vectors in the constructor and assigning it directly. It may also be a better interface for the user, they can initialize their grid directly, e.g.:
```C++
Grid grid({{'A,'B','C}, {'D','E','F'}});  // A vector of vectors via initializer list.
```

ayeletK/assignment3/position.cpp
```diff
@@ -0,0 +1,19 @@
...
+    Position(const Position& pos): _row(pos._row), _col(pos._col) {}
+    Position& operator=(const Position& pos) {
+        _row = pos._row;
+        _col = pos._col;
+        return *this;
+    }
+    bool operator<(const Position& pos) const{
+        return _row < pos._row || _col < pos._col;
+    }
+    bool equal(Position pos){
```
Brusi: You can overload `==` operator:
```C++
bool operator==(Position pos) {
    return _row == pos._row && _col == pos._col;
}
```
And then compare using the operator directly:
```C++
Position p1(1,2), p2(3,4);
return p1 == p2;
```

Unless there is a specific reason to have a function (for example. want to keep a function pointer or use it in other contexts), it is common to use this power of C++ and override the existing compare operator.

ayeletK/assignment3/position.cpp
```diff
@@ -0,0 +1,19 @@
+class Position{
```
Brusi: Nice class/struct in order to be used inside a set.

Tip:
Since all its functions and fields are public, you can define this a `struct`.

Another tip:
An alternative to your own struct can be using the `std::pair` library class of C++.
Pair holds two values, in this case `std::pair<int, int>` would do the trick.

This has both advantages and disadvantages:

Reasons to use a pair:
1. Common to all readers - users know what a pair is even if they are new to your code
2. Compare and equality operators are built in (can be used in a set directly without additional code. Even in a hash set).

Reasons to use a dedicated struct (e.g. Position):
1. Using the struct name is more readable (The intent in `set<Position>` is clearer than `set<pair<int,int>>`).
2. You have control over the field names (you could call them row/col instead of first/second), more readable.

For completion, Java language has no pair-like class for this reason exactly: they prefer a dedicated class/struct for each particular case. In C++ we have the freedom to choose :)

ayeletK/assignment3/dictionary.cpp
```diff
@@ -0,0 +1,28 @@
...
+#include <vector>
+
+using namespace std;
+
+class Dictionary{
+    set<string> _words;
+    set<string> _prefixes;
+public:
+    Dictionary(set<string> words): _words(words) {
+        vector<char> curr_prefix;
```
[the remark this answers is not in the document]
ayeletK: I did not know that. Good to know :)

ayeletK/assignment3/grid.cpp
```diff
@@ -0,0 +1,120 @@
+#include "position.cpp"
+#include "dictionary.cpp"
+
+class Grid{
+    int _rows, _cols;
+    vector<vector<char>> _grid;
+    bool update_location(int* row, int* col){
```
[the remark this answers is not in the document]
ayeletK: I wanted to define my own Location/Position class and write an iterator for it. But I had to know the grid dimensions for it so I got stuck.
I will add a comment this time. Next time I will try to come up with a better implementation.

ayeletK/assignment3/grid.cpp
```diff
@@ -0,0 +1,120 @@
...
+        if (*col + 1 < _cols){ 
+            (*col)++;
+            return true;
+        }
+        else if (*row + 1 < _rows) {
```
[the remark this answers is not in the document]
ayeletK: True.

ayeletK/assignment3/grid.cpp
```diff
@@ -0,0 +1,120 @@
...
+        int row = 0, col = 0;
+        for (int i = 0; i < data.size(); ++i){
+            if (col == 0){
+                _grid.push_back(vector<char>());
+            }
+            _grid.at(row).push_back(data.at(i));
+            if (!update_location(&row, &col)) break;
+        }
+    }
+    void update_value(int row, int col, char val){
```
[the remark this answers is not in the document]
ayeletK: I prefer the "less functions" idea :)

ayeletK/assignment3/grid.cpp
```diff
@@ -0,0 +1,120 @@
...
+        if (legal_row(row) && legal_col(col))
+            _grid.at(row).at(col) = val;
+    }
+    char get_value(int row, int col){
+        if (legal_row(row) && legal_col(col))
+            return _grid.at(row).at(col);
+    }
+    char get_value(Position pos){
```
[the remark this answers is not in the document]
ayeletK: You are right. I missed that case. I will fix it!

ayeletK/assignment3/grid.cpp
```diff
@@ -0,0 +1,120 @@
...
+        if (legal_position(pos))
+            return _grid.at(pos._row).at(pos._col);
```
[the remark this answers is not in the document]
ayeletK: I will change it. Thanks!

ayeletK/assignment3/grid.cpp
ayeletK: 👍

ayeletK/assignment3/grid.cpp
ayeletK: Thank you :)

ayeletK/assignment3/grid.cpp
ayeletK: I had a hard time with the initialisation - I wanted the "user" to have as much freedom as possible. I tend to over think things a lot :/
I like your way better. It looks clearer. 
Regarding the new constructor suggestion - I thought about it. Going this way means I have to verify the user provided a legal grid (the dimensions of each row are the same).

ayeletK/assignment3/position.cpp
ayeletK: Ok! I'll fix it.

ayeletK/assignment3/position.cpp
ayeletK: What about a combination of them both?
```c++
class Position{
  pair<int,int> _pos;
public:
  Position(int row, int col): pos(row,col) {}
  int get_row(){ return _pos.first; }
  int get_col(){ return _pos.second; }
  bool operator==(Position pos) { return this->_pos == pos._pos; }
  bool operator<(Position pos) { return this->_pos < pos._pos; }
};
```

ayeletK/assignment3/grid.cpp
ayeletK: True. as we discussed in the online session :)

ayeletK/assignment3/grid.cpp
ayeletK: Noted. i will leave it in a set this time.

PR #22, opened 2018-06-23
Assignment 3 solution

natiami:


natiami/assignment3/FindWordsTest.cpp
```diff
@@ -0,0 +1,51 @@
+#include "FindWordsTest.h"
+
+void FindWordsTest::test(){
+	test_findAllWords();
+	test_EmptyGrid();
+	test_EmptyDictionary();
+}
+
+Grid* FindWordsTest::initGrid(){
```
Brusi: Tip 1:
This can be replaced with a one liner, using vector initializer list:
```C++
return new Grid{{'A', 'A', 'R'}, {'T', 'C', 'D'}};
```
Explaination:
A vector can be initialized with a list using braces initialization, for example: `vector<int> v{1,2,3};`
Here you have a vector of vectors, so each element of the vector is a vector itself, so you can use nested braces.

-----
Tip 2:
If you dynamically allogate a grid (using `new` operator) and return it by pointer (`grid*`), then the caller should release it:
```C++
Grid* grid = initGrid();
// Use grid...
delete grid;
```
or there will be a memory leak!

So here you can return the Grid by value, so there is no need to deal with memory management:
```C++
Grid initGrid() { return Grid{{'A', 'A', 'R'}, {'T', 'C', 'D'}}; }
```
Or even:
```C++
Grid initGrid() { return {{'A', 'A', 'R'}, {'T', 'C', 'D'}}; }
```

Another option, is to make `initGrid` return nothing, and initialize the `grid` field of the test class:
```C++
void initGrid() {
    grid = Grid{{'A', 'A', 'R'}, {'T', 'C', 'D'}};  // grid is a field of the test class.
    // Or:
    grid = {{'A', 'A', 'R'}, {'T', 'C', 'D'}};
}
```

natiami/assignment3/FindWordsTest.cpp
```diff
@@ -0,0 +1,51 @@
...
+	Grid* grid = new Grid(3, 2);
+	for (size_t j = 0; j < grid->width; j++)
+	{
+		grid->cells[0][j] = gridRow1[j];
+		grid->cells[1][j] = gridRow2[j];
+	}
+	return grid;
+}
+
+Dictionary* FindWordsTest::initDictionary(){
```
Brusi: Same here, return Dictionary by value to avoid memory leak.

natiami/assignment3/Dictionary.cpp
```diff
@@ -0,0 +1,18 @@
+#include "Dictionary.h"
+
+Dictionary::Dictionary(){
+	this->words = set<string>();
```
Brusi: You don't have to use the `this` qualifier whenever accessing a class field.
Just `words` works too.

Some prefer to always use `this`, some prefer to use it sometimes (e.g. to distinguish from a function parameter with a similar name), and some never use it.
That's a matter of style and personal preference, you can choose the style which you thinks makes the code most clear and readable (or alternatively, the style that you are used to from previous works, that's ok too ;) )

natiami/assignment3/FindWordsTest.h
```diff
@@ -0,0 +1,16 @@
...
+class FindWordsTest{
+public:
+	void test();
+private:
+	Grid* initGrid();
+	Dictionary* initDictionary();
+	void test_findAllWords();
+	void test_EmptyGrid();
+	void test_EmptyDictionary();
+};
```
Brusi: That's a great test class structure and implementation! 👍 
Tests are separated to functions, and from the function name it is clear what is being tested.

natiami/assignment3/Grid.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+Grid::Grid(){
+	this->cells = vector<vector<char> >();
+}
+
+Grid::Grid(int width, int height){
+	this->width = width;
+	this->height = height;
+	this->cells = vector<vector<char> >(height, vector<char>(width));
+	this->moves = { 0, -1, -1, 0, 0, 1, 1, 0, -1, -1, -1, 1, 1, -1, 1, 1 };
```
Brusi: Tip:

This flat list of integers can be to follow, especially if you want to update it. It is actually a list of pairs! So for this you can use the std::pair class:
```C++
std::vector<std::pair<int, int>> moves;
// Then to initialize the moves you can still use braces initialization:
moves = { {0, -1}, {-1, 0}, {0, 1}, {1,0} ...};  // The intent is clearer.
// Then the iteration is easier. No need to deal with iterating by jumps of 2 etc.
for (const std::pair<int, int>& move : moves) {
   int next_i = i + move.first;
   int next_j = j + move.second;
}

```

natiami/assignment3/FindWordsTest.cpp
```diff
@@ -0,0 +1,51 @@
...
+	Dictionary* dictionary = initDictionary();
+	Grid* grid = initGrid();
+	set<string> found_words = grid->findAllWords(dictionary);
+	set<string> expected_words = { "CAR", "CARD", "CAT" };
+	EXPECT_EQ(expected_words, found_words);
+}
+
+void FindWordsTest::test_EmptyGrid(){
+	Dictionary* dictionary = initDictionary();
+	Grid* grid = new Grid();
```
Brusi: Tip:
You can initialize the Grid by value, and give it a meaningful name:
```
Grid emptyGrid;
set<string> empty_expected_words;
EXPECT_EQ(empty_expected_words, empty_grid->findAllWords(dictionary));
```
This way this test is read like a sentence:
"Expect an empty result if searching the empty grid".

Also there is would be no memory leak of the grid.

PR #23, opened 2018-06-27
Assignment 4 solution

michbl:


michbl/assignment4/assignment4.cpp
```diff
@@ -0,0 +1,126 @@
+#include <iostream>
+#include "codeu_test_lib.h"
+
+#define LAND true
+#define SEA false
```
Brusi: You can define global constants instead of using the preprocessor's `#define`:
```C++
const bool LAND = true;
const bool SEA = false;
```
In C++ using constants is the recommended way to define global constant values. They can be scoped (defined in a namespace or a class, e,g, MyClass::LAND), and can give better compile time errors.

Another option is to use enum as you have used for the Marking.

If you want to keep it `bool` (for example if the interface must be a grid of bool), you can give the grid a name which better represents the meaning of its content, e.g. `isLand` instead of `grid`. This way the boolean expressions can be more meaningful:
```C++
if (isLand[row][col]) { /* Mark the land */ }
```

michbl/assignment4/assignment4.cpp
```diff
@@ -0,0 +1,126 @@
...
+    resetGrid(rows, cols, grid);
+    grid[0][0] = grid[0][2] = grid[1][1] = grid[1][3] = grid[2][0] = grid[2][2] = grid[3][1] = grid[3][3] = true;
+    printGrid(rows, cols, grid);
+    EXPECT_EQ(countIslands(rows, cols, grid), 8);
+
+    //Test 7 - 1 land tile
+    resetGrid(rows, cols, grid);
+    grid[3][3] = true;
+    printGrid(rows, cols, grid);
+    EXPECT_EQ(countIslands(rows, cols, grid), 1);
```
Brusi: Very good tests, seem to cover many edge cases.

What about testing an empty grid?
(When you think of edge cases, always try the "empty" case. An empty string, an empty array, an empty grid...).

michbl:
I read your comments, I agree about the vectors, thanks for that!
As for a different algorithm, I'd probably use a Union-Find data structure and go over the tiles once (I think).
It's my exams period now so I'm focusing on studying, so I won't upload a fixed version this time

Brusi:
No problem, as everything post-approval is optional :)

PR #24, opened 2018-06-29
assignment 4 solution

natiami:


natiami/assignment4/FindIslands.cpp
```diff
@@ -0,0 +1,45 @@
+#include"FindIslands.h"
+
+FindIslands::FindIslands(){}
+
+FindIslands::FindIslands(int nr_row, int nr_column, vector<vector<bool> >& mapOfTiles) : nr_row(nr_row), nr_column(nr_column), mapOfTiles(mapOfTiles){
```
Brusi: The number of rows and columns can be deduced from the `mapOfTiles` parameter.
Using the actual dimensions of `mapOfTiles` both makes the interface simpler (less constructor arguments) and can be safer (no danger of a mismatch between the `nr_row`, `nr_column` and the actual dimensions of the tiles map.

To be extra safe, you can add a check that the dimensions fit or that `mapOfTiles` have all its rows in the same length (optional).

natiami/assignment4/FindIslands.cpp
```diff
@@ -0,0 +1,45 @@
+#include"FindIslands.h"
+
+FindIslands::FindIslands(){}
+
+FindIslands::FindIslands(int nr_row, int nr_column, vector<vector<bool> >& mapOfTiles) : nr_row(nr_row), nr_column(nr_column), mapOfTiles(mapOfTiles){
+}
+
+bool FindIslands::checkBounds( int i, int j){
```
Brusi: Tip:
`checkBounds` is a good name, however you can give it a name that can be more read as a sentence, like `isInBounds` or `inBounds`:
```C++
if (inBounds(i,j)) {  // Read like a sentence: "if point (i,j) is in bounds then ... "
   // Do something with
}
```

natiami/assignment4/FindIslands.cpp
```diff
@@ -0,0 +1,45 @@
...
+	visited[i][j] = true;
+	vector<pair<int, int> > moves{ { 0, 1 }, { 1, 0 }, { 0, -1 }, { -1, 0 } };
+	int row, column;
+
+	for (size_t k = 0; k < moves.size(); k++)
+	{
+		row = i + moves[k].first;
+		column = j + moves[k].second;
+		if (checkBounds(row, column)){
+			if (!visited[row][column] && mapOfTiles[row][column])
```
Brusi: The line
```C++
if (!visited[row][column] && mapOfTiles[row][column])
```
is duplicated both in here and in `findNumberOfIsland`.

To avoid this duplication you can move these checks it to the beginning of `visitOneIsland`:
```C++
void FindIslands::visitOneIsland(vector<vector<bool> >& visited, int i, int j){
    if (!checkBounds(i,j) || !mapOfTiles[i][j] || visited[i][j]) {
        return;
    }
    ...
    for (...) {
        row=...;
        col=...;
        visitOneIsland(visited, row, col);  // You can now call this directly from the loop without surrounding with checks.
    }
}
```
Same below - `visitOneIsland` can be called directly from the loop in `findNumberOfIslands`.

natiami/assignment4/FindIslands.cpp
```diff
@@ -0,0 +1,45 @@
...
+bool FindIslands::checkBounds( int i, int j){
+	return i >= 0 && i < nr_row && j >= 0 && j < nr_column;
+}
+
+void FindIslands::visitOneIsland(vector<vector<bool> >& visited, int i, int j){
+	visited[i][j] = true;
+	vector<pair<int, int> > moves{ { 0, 1 }, { 1, 0 }, { 0, -1 }, { -1, 0 } };
+	int row, column;
+
+	for (size_t k = 0; k < moves.size(); k++)
```
Brusi: Tip:
A "foreach" loop can fit here:
```C++
for (const std::pair<int, int>& move : moves) {
    row = i + move.first;
    col = j + move.second;
    ...
}
```

natiami/assignment4/FindIslands.h
```diff
@@ -0,0 +1,16 @@
+#include<vector>
+using namespace std;
+
+class FindIslands{
+
+public:
+	int nr_row;
+	int nr_column;
+	vector<vector<bool> > mapOfTiles;
```
Brusi: Tip:
You can replace `mapOfTiles` with a name that better represents the meaning of the content of the map.
What "true" and "false" mean here? True means land, False mean sea.
So `isLand` or `landTiles` or `isLandTile` can make the code read more like English:
```
if (isLandTile[i][j]) {  // "if (i,j) is a land tile then... "
   ...
}
```

natiami/assignment4/FindIslandsTest.cpp
```diff
@@ -0,0 +1,61 @@
...
+
+void FindIslandsTest::test_findNumberOfIslands(){
+	vector<vector<bool> > mapOfTiles{ { false, true, false, true },
+								{ true, true, false, false },
+								{ false, false, true, false },
+								{ false, false, true, false } };
+	findIslands = FindIslands(4, 4, mapOfTiles);
+
+	int expected_nr_island = 3;
+	EXPECT_EQ(expected_nr_island, findIslands.findNumberOfIslands());
```
Brusi: Tip:
If you want, you can make it an (almost) one-liner:
```C++
vector<vector<bool>> mapOfTiles = { ... } ;
EXPECT_EQ(3, FindIslands(mapOfTiles).findNumberOfIslands());  // <-- This is the "one liner".
```

The `EXPECT_EQ(expected, actual)` pattern is very common, so I think it is clear enough that the `3` on the left is the expected result.
However for more complex statements (which does not fit in a line) it is also very clear to put the expected result in a variable named "expected_..." and then comparing it, exactly as you did; so both options are good. Your choice :)

natiami/assignment4/FindIslands.cpp
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+}
+
+int FindIslands::findNumberOfIslands(){
+
+	vector<vector<bool> > visited(nr_row, vector<bool>(nr_column));
+	int nr_island = 0;
+
+	for (size_t i = 0; i < nr_row; i++)
```
Brusi: Note that `size_t` is an unsigned int type.
The [Google Stype Guide](https://google.github.io/styleguide/cppguide.html#Integer_Types) recommends avoiding unsigned ints, unless you need the extra values space, even if you are sure that the number is always non-negative.
The main reason is that the compiler implicitly converts signed to unsigned ints and vice versa, overflowing the values, which can cause errors that are easy to miss. For example:
```C++
size_t x = -1;
bool b = x > 0;  // Evaluates to true!!
cout << x; // Prints: 18446744073709551615
```

More info: http://wesmckinney.com/blog/avoid-unsigned-integers/

In this case, you even compare it with `nr_row` which is an int, so it is best to just keep using ints.

PR #25, opened 2018-06-30
Tijana sonjic 4

TijanaSonjic:
Solution for assignment number 4.

### DIFF
--- a/TijanaSonjic/assignment4/test.cpp
+++ b/TijanaSonjic/assignment4/test.cpp
@@ -1,0 +1,43 @@
+#include "IslandCount.h"
+#include "test_library.h"
+#include <iostream>
+
+using namespace std;
+
+void testEqual(IslandCount* islandCount, int num) {
+	EXPECT_EQ(islandCount->countIslands(), num);
+}
+
+int main() {
+	bool** map = new bool*[4];
+	map[0] = new bool[4]{ false, true, false, true };
+	map[1] = new bool[4]{ true, true, false, false };
+	map[2] = new bool[4]{ false, false, true, false };
+	map[3] = new bool[4]{ false, false, true, false };
+	
+	IslandCount* islandCount = new IslandCount(map, 4, 4);
+
+	testEqual(islandCount, 3);
+
+	
+	// Test with null map
+	IslandCount* islandCountWithoutMap = new IslandCount(nullptr, 0, 0);
+
+	testEqual(islandCountWithoutMap, 0);
+
+
+	// Test with one island across th whole map
+	bool** mapWithOneIsland = new bool*[1];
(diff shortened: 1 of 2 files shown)


TijanaSonjic/assignment4/test.cpp
```diff
@@ -0,0 +1,43 @@
+#include "IslandCount.h"
+#include "test_library.h"
+#include <iostream>
+
+using namespace std;
+
+void testEqual(IslandCount* islandCount, int num) {
+	EXPECT_EQ(islandCount->countIslands(), num);
+}
```
Brusi: Functions are good to encapsulate some logic under a function (labeled with a meaningful name).
Another benefit of functions is to avoid code duplication by extracting it to a function and calling the function several times, instead of copying the code per use.
A downside of functions is that it "hides" its content from the user, which has to go look for the its implementation if they want to see what it does.

Here, I think the function implementation is trivial, so the downsides of putting this in a function exceeds the benefits.
This function is a very simple and clear one-liner, and writing explicitly in the code several times is the most readable in my opinion:
```C++
...
EXPECT_EQ(islandCount->countIslands(), 3);
...
EXPECT_EQ(islandCountWithoutMap->countIslands(), 0);
...
EXPECT_EQ(mapWithOneIsland->countIslands(), 1);
```

Another specific downside of extracting putting `EXPECT_EQ` macro in a function, is less meaningful error messages:
When `EXPECT_EQ` failes (i.e. getting not-equal arguments), it prints the line of code in which the macro was called. When calling it explicitly, the line would direct you to the failing test case.
When calling it in a function, you will always get the same line as error, making it harder to see what exactly went wrong.
(For code which should be maintained for a long time, good tests with meaningful error messages are very important)

TijanaSonjic/assignment4/test.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+int main() {
+	bool** map = new bool*[4];
+	map[0] = new bool[4]{ false, true, false, true };
+	map[1] = new bool[4]{ true, true, false, false };
+	map[2] = new bool[4]{ false, false, true, false };
+	map[3] = new bool[4]{ false, false, true, false };
+	
+	IslandCount* islandCount = new IslandCount(map, 4, 4);
```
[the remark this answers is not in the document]
TijanaSonjic: Right, thank you.

TijanaSonjic/assignment4/test.cpp
TijanaSonjic: Yes, with this function I ruin the point of EXPECT_EQ macro.  I missed this out because the code is moved to the function after I have finished testing my code with different parameters.

TijanaSonjic:
Soon, I will submit an iterative solution.

PR #26, opened 2018-07-01
Solution for Assignment 4

annnufan:
Ready for first iteration

### DIFF
--- a/annnufan/assignment4/islands_counter.cpp
+++ b/annnufan/assignment4/islands_counter.cpp
@@ -3,52 +3,40 @@
 
 class islands_counter {
 	std::vector<std::vector<bool>> island_map;
-	std::vector<std::pair<int, int>> moves;
+	std::vector<std::pair<int, int>> moves = {{-1, 0}, {1, 0}, {0, -1}, {0, 1}};
 	int n, m;
-	int islands;
-	bool is_count;
+	int island_count;
+	bool already_count;
 
-	void dfs(int i, int j) {
+	void mark_island(int i, int j) {
 		island_map[i][j] = false;
 
 		for (const auto& move : moves) {
 			int i1 = move.first + i;
 			int j1 = move.second + j;
 			if (0 <= i1 && i1 < n && 0 <= j1 && j1 < m && island_map[i1][j1]) {
-				dfs(i1, j1);
+				mark_island(i1, j1);
 			}
 		}
 	}
 
 public:
-	islands_counter(std::vector<std::vector<bool>> i_m) : island_map(i_m) {
-		n = island_map.size();
-		m = (n != 0) ? island_map[0].size() : 0;
-		islands = 0;
-		is_count = false;
-		for (int i = -1; i <= 1; i++) {
-			for (int j = -1; j <= 1; j++) {
-				if ((i == 0) ^ (j == 0)) {
-					moves.push_back({i, j});
-				}
-			}
-		}
-	}
+	islands_counter(std::vector<std::vector<bool>>& i_m) : island_map(i_m), n(island_map.size()), m((n != 0) ? island_map[0].size() : 0), island_count(0), already_count(false) {}
 
 	int count() {
-		if (is_count) {
-			return islands;
+		if (already_count) {
+			return island_count;
 		}
-		is_count = true;
+		already_count = true;
 		for (int i = 0; i < n; i++) {
 			for (int j = 0; j < m; j++) {
 				if (island_map[i][j]) {
-					dfs(i, j);
-					islands++;
+					mark_island(i, j);
+					island_count++;
 				}
 			}
 		}
-		return islands;
+		return island_count;
 	}
 
 };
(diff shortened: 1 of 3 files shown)


annnufan/assignment4/islands_counter.cpp
```diff
@@ -0,0 +1,59 @@
+#include <vector>
+#include <utility>
+
+class islands_counter {
+	std::vector<std::vector<bool>> island_map;
+	std::vector<std::pair<int, int>> moves;
+	int n, m;
+	int islands;
+	bool is_count;
```
Brusi: Naming tips:

A reader may confuse the variable `islands` with `islands_map`. I think `num_islands` or `island_count` can make it clear and distinguishable and meaningful.

`is_count` is not intuitive to understand unless you see where it is used.
You can replace it with `already_count` or even `num_islands_already_calculated` or `has_result`.
Think of it as a part of an English sentence. 
`if (has_result)` or `if (already_count)` can read as a sentence and are clear for the reader.
Even long names are OK if they help readability.

annnufan/assignment4/test_islands_counter.cpp
```diff
@@ -0,0 +1,41 @@
...
+#include <iostream>
+
+void test_for_example() {
+	std::vector<std::vector<bool>> ex_map({{false, true, false, true}, {true, true, false, false}, {false, false, true, false}, {false, false, true, false}});
+	int ans = 3;
+	EXPECT_EQ(count_islands_on_map(ex_map), ans);
+}
+
+void test_empty_case() {
+	std::vector<std::vector<bool>> empty_map({});
```
Brusi: There is also the case of a maps with empty rows you can check:
`{{},{},{}}`

annnufan/assignment4/islands_counter.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+	void dfs(int i, int j) {
+		island_map[i][j] = false;
+
+		for (const auto& move : moves) {
+			int i1 = move.first + i;
+			int j1 = move.second + j;
+			if (0 <= i1 && i1 < n && 0 <= j1 && j1 < m && island_map[i1][j1]) {
+				dfs(i1, j1);
```
Brusi: My personal opinion:
I prefer names that are less "mathematical", and more related to the context.
I think `DFS` is a good name when accepting a general graph, and performing a general, mathematical DFS algorithm.
Here this DFS actually have a purpose: it markes islands as visited (more like - removes them from the map).
So I think `mark_island` or `remove_island` or `clear_island` are better names.
Then with a comment, you can specify that you implemented it with DFS (as it is an implementation detail).

This matches the pattern in which the function/class name represents its interface and not its implemntation, represents the WHAT and not the HOW.
For example, this is why in C++, a linked list is called "list", a tree is called "set" and a hash table is called "unordered_set". The 'hash table' is just an implementation detail; the actual interface is a set that has not order on the elements. 
This is from a pure "object oriented" perspective - encapsulate!

Arguments against what I just said:
1. This is a private function so you can have freedom in how you call it. This is not as important as the outer interface of the public functions
2. implementation is still important sometimes and you do not always want to completely encapsulate it. For example, everyone should know that the unordered_set is a hash table and understand the implications of it. So keep that in mind as well. The "pure OOP" is not always the right way to go, though it it probably best most of the time.

annnufan/assignment4/islands_counter.cpp
```diff
@@ -0,0 +1,59 @@
...
+			}
+		}
+	}
+
+public:
+	islands_counter(std::vector<std::vector<bool>> i_m) : island_map(i_m) {
```
Brusi: You pass here `i_m` by value, which means the entire map is copied, twice:
1. Once when passing it into the function (the parameter is copy-constructed)
2. Once when copying the parameter into the `island_map` field.

You can avoid the 1st copy by passing it by const reference:
`islands_counter(const std::vector<std::vector<bool>>& i_m)`
Here and in the functions below.

Another option is to "move" keep the parameter by value, but "move" it into the field:
```
islands_counter(std::vector<std::vector<bool>> i_m) : island_map(std::move(i_m)) ...
```
This uses C++11 move semantics. This is out of our scope so I don't want to get into it, just know that it exists if you see something like this in the future. :)

annnufan/assignment4/islands_counter.cpp
```diff
@@ -0,0 +1,59 @@
...
+					islands++;
+				}
+			}
+		}
+		return islands;
+	}
+
+};
+
+int count_islands_on_map(std::vector<std::vector<bool>> island_map) {
```
Brusi: Tip:
Encapsulating this in a function is perfectly fine (it actually IS a simple function - get the island map, return the number of island). For that matter, even the existence of this class is an implementation detail! Not bad :)

Other options (just for your information):

You can implement this as a one liner, if you choose:
```C++
return islands_counter(island_map).count();
```
Or you can remove call the one liner instead of this function in the tests:
```C++
EXPECT_EQ(islands_counter(island_map).count(), 4);
```
saves this function and still readable IMO.

All of these options, including what you did, is OK, choose what you think is most readable.

PR #27, opened 2018-07-03
Assignment 4

July-Morning:
Ready for the first iteration

### DIFF
--- a/July-Morning/assignment4/disjoint_set.cpp
+++ b/July-Morning/assignment4/disjoint_set.cpp
@@ -1,16 +1,19 @@
+#include <stdexcept>
 #include "disjoint_set.hpp"
 
-DisjointSet::DisjointSet (unsigned int init_set_num)
+DisjointSet::DisjointSet (int init_set_num)
 {
+    if (init_set_num < 0)
+        throw std::invalid_argument("received negative value");
     set_num = init_set_num;
-    rank = std::vector<unsigned int>(init_set_num);
-    parent = std::vector<unsigned int>(init_set_num);
-    for (unsigned int i = 0; i < init_set_num; i++)
+    rank = std::vector<int>(init_set_num);
+    parent = std::vector<int>(init_set_num);
+    for (int i = 0; i < init_set_num; i++)
         parent[i] = i;
 }
         
 // Find set that inlcudes i
-unsigned int DisjointSet::find (unsigned int i)
+int DisjointSet::find (int i)
 {
     if (parent[i] != i)
         parent[i] = find(parent[i]);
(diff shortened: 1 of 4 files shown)


July-Morning/assignment4/disjoint_set.cpp
```diff
@@ -0,0 +1,40 @@
...
+{
+    if (parent[i] != i)
+        parent[i] = find(parent[i]);
+    return parent[i];
+}
+        
+// Merge set that includes i and set that includes j according to ranks
+bool DisjointSet::unite (unsigned int i, unsigned int j)
+{
+    if ((i >= set_num) || (j >= set_num))
```
Brusi: Can remove parentheses (less clutter):
`if (i >= set_num || j >= set_num)`

July-Morning/assignment4/disjoint_set.hpp
```diff
@@ -0,0 +1,21 @@
+#ifndef DISJOINT_SET_HPP_
+#define DISJOINT_SET__HPP_
+
+#include <vector>
+
+class DisjointSet
```
Brusi: Very good, DisjointSet class which is portable and is not coupled with this particular solution. :+1:

July-Morning/assignment4/disjoint_set.hpp
```diff
@@ -0,0 +1,21 @@
...
+{
+    private:
+        unsigned int set_num;
+        std::vector<unsigned int> parent;
```
Brusi: Unsigned integers are discouraged when you do not need the extra positive values space.
One reason for it is that C++ (following C) implicitly converts between signed and unsigned values, which can cause overflow errors. For example:
```C++
unsigned int y = -1;
bool y_is_negative = y < 0;  // Returns false!
cout << y;  // Prints: 4294967295
```
This can cause errors that the compiler does not catch and they are hard to find at runtime.
For example, if the user initializes this class with `DisjointSet(-1);` then your class will behave unexpectedly;
But if you keep this a signed int, then you will be able to check it (check that number is positive), but even if you don't, it will probably fail with an error saying that it cannot initialize a vector with size -1, which is still can prevent the users scratching their head after this class returned 4294967295 when they expected it to return 4. :)

-----
Google style guide on unsigned ints:
https://google.github.io/styleguide/cppguide.html#Integer_Types

July-Morning/assignment4/disjoint_set.hpp
```diff
@@ -0,0 +1,21 @@
+#ifndef DISJOINT_SET_HPP_
+#define DISJOINT_SET__HPP_
```
Brusi: Not that important, but the `#ifndef` and `#define` have different strings so it is probably not going to do what you expect :)

July-Morning/assignment4/disjoint_set.hpp
July-Morning: Thank you for the review and tips! Got it about unsigned int. However could you please tell me what is more appropriate when the constructor is called with a negative set number parameter: raise an exception, add another param as a flag if initialization was successful or not, have a separate init method or silently (to user) set it to zero (and leave a comment in the class code describing such behaviour). 
From my point of view in real life it depends heavily on where and how the class is going to be used, but I would appreciate some comments from you about this particular case.

Brusi: The standard way to fail the constructor is to throw an exception. I think it is also the best thing to do when getting a negative number in a non-negative parameter.
It is better than failing silently (then the class would just behave funny later)
It is better than failing later in a funny way when trying to access the negative/sized vector.

Exceptions also do not change the interface of the constructor - no need for extra output parameters. It helps to keep the code simple.

Exceptions should be used in cases you really do not expect to happen (for example, negative vector size). For possible failures, maybe other solutions (e.g. an init function) can work.

An extra `init` class is also a very common pattern, especially if the constructor is not trivial, it also fits sometimes, and you can have use its return value to see if the class was initialized correctly. But it is still a little error prone, because the caller might forget to check `init` return value or forget to call it altogether.

PR #29, opened 2018-07-16
Tijana sonjic

TijanaSonjic:
This is one approach for the require part on assignment 5. Girls, please do the review on this code.


group/assignment5/WordDependencyGraph.h
```diff
@@ -0,0 +1,118 @@
...
+						}
+					}
+				}
+			}
+		}
+		// iteration for the last word in the dictionary
+		string word1 = dictionary[dictionary.size()-1];
+		for (char ch : word1) {
+			// if graph does not contain pair with the first element as letter ch, letter ch will be inserted with an empty children list,
+			// by just referencing it.
```
michbl: I think you need to do this not only for the last word, but make sure to insert into your graph all characters that aren't prior to others.
Your test works because the only characters that are not oredered before any other character are the ones in the last word, so you eventually find all the letters.
For this dictionary: {"abd", "abc", "^pap", "^alk", "*$llo", "car", "01o"}
the algorithm doesn't put all of the characters inside the alphabet, since, for example, '$' isn't prior to any other character and also isn't part of the last word, so it doesn't get into the alphabet.

TijanaSonjic: Yes, you are right :) Thank you! I will correct this.

group/assignment5/test1.cpp
```diff
@@ -0,0 +1,30 @@
+#include <iostream>
+#include <string>
+#include "WordDependencyGraph.h"
+#include "test_library.h"
+
+using namespace std;
+
+
+int main() {
+	// test 1
```
July-Morning: I believe more interesting test cases could be added here: empty dictionary, empty words in the dictionary, identical words, etc. In addition it's sometimes useful to make tests separate and give them clear names so that if test fails you (or not you) can see what exactly went wrong :)

group/assignment5/WordDependencyGraph.h
```diff
@@ -0,0 +1,118 @@
...
+					string word2 = dictionary[j];
+
+					if (word1.length() > s && word2.length() > s && haveTheSamePrefix(word1, word2, s)) {
+						char ch1 = word1[s];
+						char ch2 = word2[s];
+
+						if (ch1 != ch2) {
+							// ch1 should be ordered before ch2 => ch2 is dependent on ch1
+							// insert ch2 as a child of ch1 node if it is not already inserted
+							vector<char>& childrenOfChar1 = graph[ch1];
```
July-Morning: Maybe it's very subjective, but I don't like the name "childrenOfChar1" :) For me "adjacentToCurChar" or something like that would do better, but that's just a suggestion :)

group/assignment5/WordDependencyGraph.h
```diff
@@ -0,0 +1,118 @@
...
+								childrenOfChar1.push_back(ch2);
+							}
+						}
+					}
+				}
+			}
+		}
+	}
+
+	bool containsChar(vector<char>& chars, char ch2) {
```
July-Morning: These small self-explainable functions make code more structured and easy to read, cool :+1:

PR #30, opened 2018-07-16
michal's solution for assignment 5

michbl:
Hi guys,

This is my solution for the 5th assignment. It should find an alphabet and should also find inconsistencies in dictionaries (optional challenge #2)

Let me know what you think :)

### DIFF
--- a/group/assignment5/main.cpp
+++ b/group/assignment5/main.cpp
@@ -1,112 +1,8 @@
-#include <iostream>
-#include <cmath>
-#include <vector>
-#include <unordered_map>
+#include "AlphabetFromDictionary.hpp"
 #include "codeu_test_lib.h"
 
-using namespace std;
-
-vector<char> findAlphabetCharacters(const vector<string> &dictionary) {
-    vector<bool> charsInDict((unsigned int)pow(2, CHAR_BIT), false);
-    vector<char> alphabetChars;
-    for (auto word : dictionary) {
-        for (char c : word) {
-            if (!charsInDict[c]) {
-                charsInDict[c] = true;
-                alphabetChars.push_back(c);
-            }
-        }
-    }
-    return alphabetChars;
-}
-
-bool createCharPriorityGraph(vector<string> &dictionary, unordered_map<char, vector<bool>> &charPriorityGraph, unordered_map<char, int> &numOfPriorCharacters) {
-    while (dictionary.size() > 0 && dictionary[0].size() == 0) {
-        dictionary.erase(dictionary.begin());
-    }
-    if (dictionary.size() < 2) {
-        return true;
-    }
-    char prevChar = dictionary[0][0];
-    unordered_map<char, vector<string>> suffixesByCharacters;
-    suffixesByCharacters[prevChar] = vector<string>(0);
-    for (string word : dictionary) {
-        if (word.size() == 0) {
-            continue;
-        }
-        char currentChar = word[0];
-        if (currentChar != prevChar) {
-            if (suffixesByCharacters.find(currentChar) != suffixesByCharacters.end()) { // found a circle in characters - inconsistent dictionary
-                return false;
-            }
-            // mark prevChar as prior to currentChar and add new suffixes vector for current char
-            charPriorityGraph[prevChar][currentChar] = true;
-            suffixesByCharacters[currentChar] = vector<string>(0);
-            numOfPriorCharacters[currentChar]++;
-        }
-        word.erase(0, 1);
-        suffixesByCharacters[currentChar].push_back(word);
-        prevChar = currentChar;
-    }
-    for (auto it : suffixesByCharacters) {
-        vector<string> suffixes = it.second;
-        createCharPriorityGraph(suffixes, charPriorityGraph, numOfPriorCharacters);
-    }
-    return true;
-}
-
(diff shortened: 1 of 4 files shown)


group/assignment5/main.cpp
```diff
@@ -0,0 +1,194 @@
...
+        if (currentChar != prevChar) {
+            if (suffixesByCharacters.find(currentChar) != suffixesByCharacters.end()) { // found a circle in characters - inconsistent dictionary
+                return false;
+            }
+            // mark prevChar as prior to currentChar and add new suffixes vector for current char
+            charPriorityGraph[prevChar][currentChar] = true;
+            suffixesByCharacters[currentChar] = vector<string>(0);
+            numOfPriorCharacters[currentChar]++;
+        }
+        word.erase(0, 1);
```
TijanaSonjic: Erase function effectively reduces the container size, but because vectors use an array as their underlying storage, erasing elements in positions other than the vector end causes the container to relocate all the elements after the segment erased to their new positions. This way solution is more cleaner, but it would be efficient if you just track current position in the word.

group/assignment5/main.cpp
```diff
@@ -0,0 +1,194 @@
...
+        prevChar = currentChar;
+    }
+    for (auto it : suffixesByCharacters) {
+        vector<string> suffixes = it.second;
+        createCharPriorityGraph(suffixes, charPriorityGraph, numOfPriorCharacters);
+    }
+    return true;
+}
+
+bool findNextOrigin(const unordered_map<char, int> &numOfPriorCharacters, char &origin) {
```
TijanaSonjic: Wow 👍 Great way to detect if we have circles in graph, which means that we have inconsistent dictionary.

group/assignment5/main.cpp
```diff
@@ -0,0 +1,194 @@
...
+            origin = it.first;
+            return true;
+        }
+    }
+    // no origin -> circle in graph -> dictionary is inconsistent
+    return false;
+}
+
+void updateNumOfPriorCharacters(const unordered_map<char, vector<bool>> &charPriorityGraph, unordered_map<char, int> &numOfPriorCharacters, const char origin) {
+    const vector<bool> children = charPriorityGraph.at(origin);
```
TijanaSonjic: I didn't know that instead of operator[] you can use function at(int pos), which will check if index is out of bounds before it try to access the element at specified position. :) thanks, now I know the safe way to access the element of vector :)

group/assignment5/main.cpp
```diff
@@ -0,0 +1,194 @@
...
+    EXPECT_EQ(expectedChars, actualChars);
+    EXPECT_TRUE(idxs['a'] < idxs['^']);
+    EXPECT_TRUE(idxs['d'] < idxs['c']);
+    EXPECT_TRUE(idxs['p'] < idxs['a']);
+    EXPECT_TRUE(idxs['^'] < idxs['*']);
+    EXPECT_TRUE(idxs['*'] < idxs['c']);
+    EXPECT_TRUE(idxs['c'] < idxs['0']);
+
+    return 0;
+}
```
TijanaSonjic: Great solution! I like tests, they are very extensive. I can suggest only to encapsulate the logic inside of the class and provide only necessary interface for users. :)

group/assignment5/main.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+int main() {
+    vector<string> dictionary = {"ART", "RAT", "CAT", "CAR"};
+    vector<char> alphabet;
+    bool result;
+    unordered_map<char, int> idxs;
+    vector<bool> expectedChars;
+    vector<bool> actualChars;
+
+    // test 1 - assignment example
```
July-Morning: Point for a discussion here: I really love the way you did the testing, because there are a lot of cases in this problem in which the right answer is not unique. So you do a general validation which can be applied even if the algorithm will change. However if we are talking about this particular implementation, it is known what exactly alphabet from a set of potential right ones we will get. Maybe it should be checked that our expectations are met?

group/assignment5/main.cpp
```diff
@@ -0,0 +1,194 @@
+#include <iostream>
+#include <cmath>
+#include <vector>
+#include <unordered_map>
+#include "codeu_test_lib.h"
+
+using namespace std;
+
+vector<char> findAlphabetCharacters(const vector<string> &dictionary) {
+    vector<bool> charsInDict((unsigned int)pow(2, CHAR_BIT), false);
```
July-Morning: I love the way you avoided magical consts here!

group/assignment5/main.cpp
```diff
@@ -0,0 +1,194 @@
...
+                alphabetChars.push_back(c);
+            }
+        }
+    }
+    return alphabetChars;
+}
+
+bool createCharPriorityGraph(vector<string> &dictionary, unordered_map<char, vector<bool>> &charPriorityGraph, unordered_map<char, int> &numOfPriorCharacters) {
+    while (dictionary.size() > 0 && dictionary[0].size() == 0) {
+        dictionary.erase(dictionary.begin());
```
July-Morning: Here a dictionary provided by user is changed, I am not sure, but it may not be always acceptable.

group/assignment5/main.cpp
natiami: Not only here but dictionary.erase(dictionary.begin()); this one too

PR #31, opened 2018-07-16
natiami's solution for assignment 5

natiami:


group/assignment5/FindAlphabet.h
```diff
@@ -0,0 +1,23 @@
+#include<vector>
+#include<iostream>
+#include<set>
+#include<map>
+#include<stack>
+#include<assert.h>
```
July-Morning: Do you actually need assert or iostream here? :)

PR #32, opened 2018-07-17
Assignment 5

July-Morning:
Here's my version of solution. 
Sorry again for being late! 
I would appreciate any comments and suggestions.


group/assignment5/main.cpp
```diff
@@ -0,0 +1,94 @@
+#include <string>
+#include "test.hpp"
+#include "orgraph.hpp"
+
+std::vector<char> get_alphabet_from_dict (std::vector<std::string> dict)
+{
+    if (dict.empty())
+        return std::vector<char>();
+    std::vector<char> sym_list;
+    std::vector<int> sym_to_id (256, -1);
```
TijanaSonjic: You can add this line at the beginning of your file and avoid writing std:: all the time :) 

using namespace std;

group/assignment5/main.cpp
```diff
@@ -0,0 +1,94 @@
+#include <string>
+#include "test.hpp"
+#include "orgraph.hpp"
+
+std::vector<char> get_alphabet_from_dict (std::vector<std::string> dict)
```
TijanaSonjic: Instead of sending dictionary by value, it's enough to sent it by reference.

group/assignment5/main.cpp
```diff
@@ -0,0 +1,94 @@
...
+{
+    if (dict.empty())
+        return std::vector<char>();
+    std::vector<char> sym_list;
+    std::vector<int> sym_to_id (256, -1);
+    int sym_num = 0;
+    for (std::string word : dict)
```
TijanaSonjic: I think that here for each word copy constructor is called?
So this should be enough: for (std::string& word : dict){...}

July-Morning: You are absolutely right, I will fix that!

PR #33, opened 2018-07-18
assignment 5 solution

michbl:


group/assignment5/main.cpp
```diff
@@ -0,0 +1,96 @@
...
+    vector<string> dictionary;
+    vector<char> alphabet;
+    bool result;
+    unordered_map<char, int> idxs;
+    vector<bool> expectedChars;
+    vector<bool> actualChars;
+
+    // test 1 - assignment example
+    dictionary = {"ART", "RAT", "CAT", "CAR"};
+    AlphabetFromDictionary afd(dictionary);
```
TijanaSonjic: When I saw _afd_ in this test, by its name I could not guess what it is supposed to be. I think that it is better to have long name alphabetFromDictionary, so reader will know meaning of it.

group/assignment5/main.cpp
```diff
@@ -0,0 +1,96 @@
...
+    idxs.clear();
```
TijanaSonjic: The same comment as for _afd_ name. Maybe _idxs_ means something? It's hard to read code with this short names, because each time you need to search for its definition.

group/assignment5/AlphabetFromDictionary.cpp
```diff
@@ -0,0 +1,94 @@
...
+    vector<char> alphabetCharacters = findAlphabetCharacters();
+    this->charPriorityGraph.clear();
+    this->numOfPriorCharacters.clear();
+    for (auto c : alphabetCharacters) {
+        this->charPriorityGraph[c] = set<char>();
+        this->numOfPriorCharacters[c] = 0;
+    }
+}
+
+bool AlphabetFromDictionary::findAlphabet(vector<char> &alphabet) {
```
TijanaSonjic: Do we want to let user to call this function multiple times? If dictionary is the same, graphCreation needlessly will be called. One option is to have boolean flag which will tell us do we need to execute this function again (if dictionary is changed we can reset that flag) and in the other option we can call this function from constructor and from setDictionary function and make it private, so the user won't be able to call it by himself. We can provide only getAlfabet() function. What do you think? :)

group/assignment5/AlphabetFromDictionary.cpp
```diff
@@ -0,0 +1,94 @@
...
+    int index = 0;
+    while (index < str1.size() && index < str2.size() && str1[index] == str2[index])
+        index++;
+    return (index < str1.size() && index < str2.size()) ? index : -1;
+}
+
+vector<char> AlphabetFromDictionary::findAlphabetCharacters() {
+    vector<bool> charsInDict((unsigned int) pow(2, CHAR_BIT), false);
+    vector<char> alphabetChars;
+    for (auto word : this->dictionary) {
```
TijanaSonjic: Why do we use auto instead of string? (I really don't know). Can we write something like this for _(auto& word : this->dictionary)??_ 
If we change it to string, we can get reference to word and avoid getting words by value from dictionary.
_for (string& word : this->dictionary)_

group/assignment5/main.cpp
michbl: Sure, we can change that, I don't mind.

group/assignment5/AlphabetFromDictionary.cpp
michbl: We can use string here, it doesn't really matter. I forgot I wrote it like that, force of habbit..

group/assignment5/AlphabetFromDictionary.cpp
TijanaSonjic: Yes, definitely we can let user to change the dictionary, but I guess we don't want to let him call this function multiple times over the same dictionary? One option was to add flag and in the other one we will call that function only from constructor and from setDictionary methods.

michbl: I agree about not calling findAlphabet() multiple times.
I think I actually prefer the second option you suggested - calling findAlphabet() from the constructor and from setDictionary()

TijanaSonjic: Me too :) Lets see what other girls think

group/assignment5/main.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+int main() {
+    vector<string> dictionary;
+    vector<char> alphabet;
+    bool result;
+    unordered_map<char, int> idxs;
+    vector<bool> expectedChars;
+    vector<bool> actualChars;
+
+    // test 1 - assignment example
```
July-Morning: I still do believe that each test case should better be a separate function with a clear name. Maybe we should also print the whole alphabet in case one of the related tests fails... It's just for convenience, though.
In addition, if I replace "CAT" with empty string in this test case, I get a following error:
`main.cpp:30 Failed: Statement (idxs['R'] < idxs['C']) is false, expected true.`
It may be a bug as 'R' still must go before 'C', am I right?

group/assignment5/main.cpp
July-Morning: I agree about _afd_ and _idxs_ names, they are a bit confusing. When you read _main.cpp_ for the first time, it's not clear what kind of indexes excatly are there.

group/assignment5/AlphabetFromDictionary.cpp
```diff
@@ -0,0 +1,94 @@
...
+        if (mismatchIndex != -1) {
+            char priorChar = this->dictionary[i][mismatchIndex], followingChar = this->dictionary[i + 1][mismatchIndex];
+            this->charPriorityGraph[priorChar].insert(followingChar);
+            this->numOfPriorCharacters[followingChar]++;
+        }
+    }
+}
+
+int AlphabetFromDictionary::findFirstMismatchIndex(string str1, string str2) {
+    if (str1.size() == 0 || str2.size() == 0) {
```
July-Morning: Not 100% sure, but is this check necessary here?

group/assignment5/AlphabetFromDictionary.cpp
```diff
@@ -0,0 +1,94 @@
...
+    }
+    return true;
+}
+
+void AlphabetFromDictionary::createCharPriorityGraph() {
+    if (this->dictionary.empty()) return;
+    int mismatchIndex;
+    for (int i = 0; i < this->dictionary.size()-1; i++)
+    {
+        mismatchIndex = findFirstMismatchIndex(this->dictionary[i], this->dictionary[i+1]);
```
July-Morning: Sometimes you write _i + 1_ and sometimes _i+1_. I prefer the first option :)

group/assignment5/AlphabetFromDictionary.cpp
July-Morning: I think for this assignment calling findAphabet only once for the dictionary is just ok, so I agree with you :)

group/assignment5/main.cpp
TijanaSonjic: I like how tests in your assignment are organized, we can do the same thing here?

July-Morning: Sure, if everyone agree, I can take care of that.

group/assignment5/AlphabetFromDictionary.cpp
TijanaSonjic: If we send empty string it want fail if we don't have this check. We can not assign null to the string in c++ so there is no need to check either if one of those strings are null.

group/assignment5/AlphabetFromDictionary.cpp
natiami: I don't agree, first if dictionary is the same and flag tells that, where is the alphabet? it is not saved in the class so it should find it again anyways. Even if you pass it to setDictionary or constructor you are not saving it in the class so how can you return it to user whenever she needs it? which will be at setDictionary? How can you obtain alphabet from class when you are not changing dictionary?

group/assignment5/AlphabetFromDictionary.cpp
```diff
@@ -0,0 +1,94 @@
...
+        if (it.second == 0) {
+            origin = it.first;
+            return true;
+        }
+    }
+    // no origin -> circle in graph -> dictionary is inconsistent
+    return false;
+}
+
+bool AlphabetFromDictionary::topologicalSort(vector<char> &alphabet) {
```
natiami: Solution finds topological sorting but it's not the algorithm of topological sort, I think complexity will improve if we use algorithm because it goes in each node once whereas this solution searches through all chars to find next char updates their priors and erases all the chars. Or we can just sort the map numOfPriorCharacters by second element. For optional #2 we can detect cyrcle during DFS or if we sort the map then we can check if consecutive numbers does not differ by 1. But sorting takes O(nlogn), so first option is better

group/assignment5/AlphabetFromDictionary.cpp
michbl: We will save the current alphabet in the class and provide a getAlphabet() function for the user

group/assignment5/main.cpp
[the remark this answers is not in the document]
michbl: Because we don't really know the order of the characters in the alphabet, so we can't compare 2 arrays. But we can use sets for a more efficient test.

group/assignment5/main.cpp
```diff
@@ -0,0 +1,96 @@
...
+    EXPECT_EQ(4, alphabet.size());
+    for (char c : {'A', 'R', 'C', 'T'}) {
+        expectedChars[c] = true;
+    }
+    for (int i = 0; i < alphabet.size(); i++) {
+        idxs[alphabet[i]] = i;
+        actualChars[alphabet[i]] = true;
+    }
+    EXPECT_EQ(expectedChars, actualChars);
+    EXPECT_TRUE(idxs['A'] < idxs['R']);
```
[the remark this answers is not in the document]
michbl: I think it's ok to have hard-coded things in a specific test. Do you have an idea for a more generic way to check if the order of the characters is correct?
(I don't think we should compare 2 arrays because the order of the characters isn't guaranteed since some characters don't have restrictions on their position in the alphabet, so we can't assume a certain order. Also, creating an array of the expected alphabet is hard-coded too.)

group/assignment5/main.cpp
```diff
@@ -0,0 +1,96 @@
...
+    dictionary = {"ART", "RAT", "ABS"};
+    afd.setDictionary(dictionary);
+    result = afd.findAlphabet(alphabet);
+    EXPECT_FALSE(result);
+    EXPECT_EQ(0, alphabet.size());
+
+    // test 4
+    dictionary = {"ABC", "DEFG", "HIJK", "LMN", "OPQRS", "TU", "VWXYZ"};
+    afd.setDictionary(dictionary);
+    expectedChars = vector<bool>(pow(2, CHAR_BIT), false);
```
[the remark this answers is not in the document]
michbl: I agree separating the tests to functions will look better and make them clearer and easier to use, but why do you say these actions aren't good?

group/assignment5/main.cpp
```diff
@@ -0,0 +1,96 @@
...
+    EXPECT_EQ(0, alphabet.size());
+
+    // test 3 - inconsistent dictionary
+    dictionary = {"ART", "RAT", "ABS"};
+    afd.setDictionary(dictionary);
+    result = afd.findAlphabet(alphabet);
+    EXPECT_FALSE(result);
+    EXPECT_EQ(0, alphabet.size());
+
+    // test 4
```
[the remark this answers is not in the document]
michbl: The test checks exactly the requirement of the assignment, which is the order of the characters that can be concluded from the dictionary. The order of the other characters is not known and doesn't matter for the test. But you're right about it not testing anything new, just another test..

group/assignment5/AlphabetFromDictionary.hpp
```diff
@@ -0,0 +1,35 @@
...
+class AlphabetFromDictionary {
+private:
+    vector<string> dictionary;
+    unordered_map<char, set<char>> charPriorityGraph;
+    unordered_map<char, int> numOfPriorCharacters;
+
+public:
+    AlphabetFromDictionary(vector<string> dictionary) : dictionary(dictionary) {};
+    ~AlphabetFromDictionary() = default;
+    bool findAlphabet(vector<char> &alphabet);
```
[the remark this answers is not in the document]
michbl: I agree, but after we make the changes of saving the alphabet in a class variable and adding a getAlphabet() function, this problem won't exist anymore.

group/assignment5/main.cpp
```diff
@@ -0,0 +1,96 @@
+#include "AlphabetFromDictionary.hpp"
+#include "codeu_test_lib.h"
+
+int main() {
```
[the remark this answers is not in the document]
michbl: Don't know if a class for testing is necessary, but I definitely agree it will be better to separate the tests into functions

group/assignment5/main.cpp
```diff
@@ -0,0 +1,96 @@
...
+    afd.setDictionary(dictionary);
+    result = afd.findAlphabet(alphabet);
+    EXPECT_FALSE(result);
+    EXPECT_EQ(0, alphabet.size());
+
+    // test 4
+    dictionary = {"ABC", "DEFG", "HIJK", "LMN", "OPQRS", "TU", "VWXYZ"};
+    afd.setDictionary(dictionary);
+    expectedChars = vector<bool>(pow(2, CHAR_BIT), false);
+    actualChars = vector<bool>(pow(2, CHAR_BIT), false);
```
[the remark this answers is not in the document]
michbl: Agree about the code repetition. About the bool vector - you're right, I just thought it will be more readable that way

group/assignment5/AlphabetFromDictionary.cpp
michbl: I think we should ignore empty strings in the dictionary since they don't add any information about the order of the characters.

group/assignment5/main.cpp
michbl: The test fails because the algorithm checks only 2 adjacent words in the dictionary at a time, and really doesn't handle empty strings (my bad :( ). I thought about 2 solutions:
a.  Erase empty strings from the dictionary.
b. Save the last non-empty string in a variable, so we don't lose information when we encounter an empty string.

What do you think? Any other ideas?

group/assignment5/AlphabetFromDictionary.cpp
michbl: I agree, let's use the known solution and use DFS for challenge 2

group/assignment5/main.cpp
July-Morning: I think we can just skip them while looping through the dictionary, which is more or less equivalent to your b option.

group/assignment5/AlphabetFromDictionary.cpp
July-Morning: Agree, from my point of view empty strings in the dictionary are acceptable.

group/assignment5/main.cpp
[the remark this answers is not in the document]
July-Morning: Code repetition should better be removed, but explicit initialization vector to "false" or zeros significantly improves readability (at least for me :) ).

PR #34, opened 2018-07-20
class corrections according to review

michbl:
I looked at all of your comments and tried fixing the solution according to them (hope I didn't miss any).
The main file isn't really different because I understand someone else will rewrite the tests, but you can look in it to see how the class is used.
It's simple (I think): initialize an instance with a dictionary (or set a dictionary with setDictionary() to an existing instance), use getAlphabet() to get the alphabet, use isDictionaryConsistent() to find if the dictionary is consistent.
Cycles in the graph are found during the creation of the graph, I think it's efficient since I add an O(1) action to each iteration, but I'm happy to hear your opinion about it.
Another thing I did and I wanted to ask your opinion about - when I find that the dictionary is inconsistent, I set the alphabet to be the characters of the dictionary in an arbitrary order. In the previous solution, I returned an empty alphabet. What do you think?

New commit - a full solution with all 3 challenges :)


group/assignment5/AlphabetFromDictionary.cpp
```diff
@@ -0,0 +1,120 @@
...
+    this->charPriorityGraph.clear();
+    for (auto c : this->alphabetCharacters) {
+        this->charPriorityGraph[c] = vector<bool>(NUM_OF_NATIVE_CHARS, false);
+    }
+}
+
+void AlphabetFromDictionary::findAlphabetCharacters() {
+    this->alphabetCharacters.clear();
+    vector<bool> isCharInDictionary(NUM_OF_NATIVE_CHARS, false);
+    for (auto word : this->dictionary) {
```
TijanaSonjic: Get each word by reference :)
for (string& word : this->dictionary)

group/assignment5/AlphabetFromDictionary.hpp
```diff
@@ -0,0 +1,41 @@
...
+    explicit AlphabetFromDictionary(const vector<string> &dictionary);;
+    ~AlphabetFromDictionary() = default;
+    void setDictionary(const vector<string> &dictionary);
+    vector<char> getAlphabet();
+    bool isDictionaryConsistent();
+
+private:
+    void initializeClassVars();
+    void findAlphabetCharacters();
+    int findFirstMismatchIndex(string str1, string str2);
```
TijanaSonjic: int findFirstMismatchIndex(const string& str1, const string& str2);
Adding the const qualifier to a reference we say that the code promises not to alter the contents of the object being referenced.

group/assignment5/AlphabetFromDictionary.cpp
```diff
@@ -0,0 +1,120 @@
...
+        for (char c : word) {
+            if (!isCharInDictionary[c]) {
+                isCharInDictionary[c] = true;
+                this->alphabetCharacters.push_back(c);
+            }
+        }
+    }
+}
+
+int AlphabetFromDictionary::findFirstMismatchIndex(string str1, string str2) {
```
TijanaSonjic: int AlphabetFromDictionary::findFirstMismatchIndex(const string& str1, const string& str2)

group/assignment5/codeu_test_lib.h
```diff
@@ -0,0 +1,80 @@
...
+template <class T>
+inline std::ostream& operator<<(std::ostream& os, const std::vector<T>& v) {
+  os << "{";
+  for (int i = 0; i < v.size(); i++) {
+    if (i > 0) {
+      os << ", ";
+    }
+    os << v[i];
+  }
+  os << "}";
```
natiami: I'm getting error here: 'operator<<<bool>' : must return a value. return os; should be added

PR #35, opened 2018-07-20
Natiami group

natiami:
Solution of challenge 1, Test cases should be added I have not done it. Also I added declaration of function for challenge 3


group/assignment5/AlphabetFromDictionary.cpp
```diff
@@ -1,120 +1,175 @@
...
-        for (char c : word) {
-            if (!isCharInDictionary[c]) {
-                isCharInDictionary[c] = true;
-                this->alphabetCharacters.push_back(c);
-            }
-        }
-    }
+	this->alphabetCharacters.clear();
+	vector<bool> isCharInDictionary(NUM_OF_NATIVE_CHARS, false);
+	for (auto word : this->dictionary) {
```
michbl: I guess that when you pressed (ctrl+k+d) it turned "string &word" back to "auto word". It doesn't really bother me but it is important to @TijanaSonjic.. Could you change it back?

group/assignment5/AlphabetFromDictionary.cpp
```diff
@@ -1,120 +1,175 @@
...
+		}
+	}
 }
 
-int AlphabetFromDictionary::findFirstMismatchIndex(const string &str1, const string &str2) {
-    int index = 0;
-    while (index < str1.size() && index < str2.size() && str1[index] == str2[index])
-        index++;
-    return (index < str1.size() && index < str2.size()) ? index : -1;
+int AlphabetFromDictionary::findFirstMismatchIndex(string &str1, string &str2) {
```
michbl: Is there a reason you remeved the "const" from the function declaration or was it (ctrl+k+d)?

group/assignment5/AlphabetFromDictionary.cpp
```diff
@@ -1,120 +1,175 @@
...
+		allAlphabets.insert(alphabet);
+}
+
+
+// TODO: challenge 1
+set<vector<char> > AlphabetFromDictionary::findAllAlphabets(){
+	set<vector<char> > allAlphabets;
+	if (dictionary.empty()) return allAlphabets;
+
+	vector<bool> visited(NUM_OF_NATIVE_CHARS);
```
michbl: I still think it's more readable to write "vector visited(NUM_OF_NATIVE_CHARS**, false**);"
Also, I think consistency in the code is important, so either we write "false" in all those declarations or we don't write it in any of them
What do you think?

group/assignment5/AlphabetFromDictionary.cpp
```diff
@@ -1,120 +1,175 @@
...
-        this->charPriorityGraph[c] = vector<bool>(NUM_OF_NATIVE_CHARS, false);
-    }
+	this->findAlphabetCharacters();
+	this->isGraphCyclic = false;
+	this->orderedAlphabet.clear();
+	this->charPriorityGraph.clear();
+	for (auto c : this->alphabetCharacters) {
+		this->charPriorityGraph[c] = vector<bool>(NUM_OF_NATIVE_CHARS, false);
+	}
+	this->inDegree.resize(NUM_OF_NATIVE_CHARS, 0);
```
michbl: When you call resize with a value (0), it adds the number of needed elements, and only the added elements receive the specified value (0).
That means that when we set a new dictionary and the size of inDegree is already 256, this line does nothing and the vector is not initialized, and it accumulates the values from previous runs.
One way to fix that is to replace this line with "this->inDegree = vector<int>(NUM_OF_NATIVE_CHARS, 0);".
This way all the elements in inDegree are 0 after initializeClassVars() is called.

group/assignment5/AlphabetFromDictionary.cpp
```diff
@@ -1,120 +1,175 @@
...
+		mismatchIndex = findFirstMismatchIndex(lastNonEmptyWord, word);
+		if (mismatchIndex != -1) {
+			char priorChar = lastNonEmptyWord[mismatchIndex], followingChar = word[mismatchIndex];
+			if (charPriorityGraph[followingChar][priorChar]) { // cycle in graph
+				this->isGraphCyclic = true;
+				this->orderedAlphabet = this->alphabetCharacters;
+				return;
+			}
+			this->charPriorityGraph[priorChar][followingChar] = true;
+			this->inDegree[followingChar]++;
```
michbl: There should be an if here:
_if (!this->charPriorityGraph[priorChar][followingChar]) {
    this->inDegree[followingChar]++;
}_
Otherwise we can count inDegree twice for the same priorChar. For example, in this dictionary: {"ART", "RAT", "CA", "CR"}, we will see that A comes before R twice, and inDegree of R will be 2 instead of 1

PR #37, opened 2018-07-23
Group 5 tests added

July-Morning:
Hope I've done everything right with branches and pull requests :)
Here's a test class with a number of test cases that more or less cover functionality of the code.
Please let me know what you think.


group/assignment5/codeu_test_lib.h
```diff
@@ -0,0 +1,94 @@
...
+template <class T>
+inline std::ostream& operator<<(std::ostream& os, const std::set<T>& s) {
+  os << "{";
+  for (auto it = s.begin(); it != s.end(); ++it)
+  {
+    if (it != s.begin())
+      os << ", "; 
+    os << *it;
+  }
+  os << "}";
```
natiami: This gives me an error as I already mentioned and I added: return os; just like it was in ori's file

group/assignment5/TestAlphabetFromDictionary.cpp
```diff
@@ -0,0 +1,161 @@
...
+    EXPECT_TRUE(alphabetFromDictionary.isDictionaryConsistent());
+    vector<char> alphabet = alphabetFromDictionary.getAlphabet();
+    set<char> actualChars(alphabet.begin(), alphabet.end());
+    set<char> expectedChars( {'A', 'R', 'C'});
+    unordered_map<char, int> charSequentialNumber;
+    EXPECT_EQ(3, alphabet.size());
+    for (int i = 0; i < alphabet.size(); i++) 
+        charSequentialNumber[alphabet[i]] = i;
+    EXPECT_EQ(expectedChars, actualChars);
+    EXPECT_TRUE(charSequentialNumber['A'] < charSequentialNumber['R']);
```
natiami: This gives false and expected true

natiami: As I wrote on Slack I think here should be:  EXPECT_TRUE(charSequentialNumber['C'] < charSequentialNumber['A']); in that case everything works well

PR #38, opened 2018-07-27
First challenge

natiami:
please review and I will merge this to group-6 branch, which you can take and make new branch from it and merge after reviews, so all the merges will be on group-6 branch and when we finish we can merge it to master. I hope this makes sense :)

### DIFF
--- a/group/assignment6/Move.h
+++ b/group/assignment6/Move.h
@@ -5,14 +5,13 @@
 struct Move
 {
 	int carId;
-	int startPosition;
-	int endPosition;
+	std::pair<int, int> positions;
 
 	Move() {}
-	Move(int carId, int startPosition, int endPosition) : carId(carId), startPosition(startPosition), endPosition(endPosition){}
+	Move(int carId, std::pair<int, int> positions) : carId(carId), positions(positions){}
 
 	friend std::ostream& operator<<(std::ostream& os, const Move& move) {
-		os << "Move car #" << move.carId << " from space " << move.startPosition << " to " << move.endPosition << std::endl;
+		os << "Move car #" << move.carId << " from space " << move.positions.first << " to " << move.positions.second << std::endl;
 		return os;
 	}
 };
--- a/group/assignment6/Move.h
+++ b/group/assignment6/Move.h
@@ -1,0 +1,20 @@
+#ifndef MOVE_H
+#define MOVE_H
+#pragma once
+
+struct Move
+{
+	int carId;
+	int startPosition;
+	int endPosition;
+
+	Move() {}
+	Move(int carId, int startPosition, int endPosition) : carId(carId), startPosition(startPosition), endPosition(endPosition){}
+
+	friend std::ostream& operator<<(std::ostream& os, const Move& move) {
+		os << "Move car #" << move.carId << " from space " << move.startPosition << " to " << move.endPosition << std::endl;
+		return os;
+	}
+};
+
+#endif
--- a/group/assignment6/RearrangeCars.cpp
+++ b/group/assignment6/RearrangeCars.cpp
@@ -1,0 +1,56 @@
+#include"RearrangeCars.h"
+
+RearrangeCars::RearrangeCars(const vector<int> &startStates) : startStates(startStates){}
+
+void RearrangeCars::setStartStates(const vector<int> &startStates){
+	this->startStates = startStates;
+}
+
+int RearrangeCars::findEmptySlotId(){
+	for (int slotId = 0; slotId < this->startStates.size(); slotId++){
+		if (this->startStates[slotId] == -1) return slotId; // no car has -1 id, it is reserved for empty slot
+	}
+	return -1; // This code should not be reached
+}
+
+vector<int> RearrangeCars::findCarPositions(const vector<int> &currentStates){
+	vector<int> carPositions(currentStates.size() - 1);
+	for (int i = 0; i < currentStates.size(); i++)
+	{
+		if (currentStates[i] != -1) // check for empty slot
+			carPositions[currentStates[i]] = i;
+	}
+	return carPositions;
+}
+
+int RearrangeCars::findMisplacedCar(const vector<int> &carPosition, const vector<int> &endStates){
+	for (int carId = 0; carId < carPosition.size(); carId++){
+		if (endStates[carPosition[carId]] != carId) return carId;
+	}
+	return -1; // If every car is on its slot
+}
+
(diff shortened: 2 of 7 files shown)


group/assignment6/RearrangeCars.h
```diff
@@ -0,0 +1,27 @@
...
+	RearrangeCars(const vector<int> &startStates);
+	~RearrangeCars() = default;
+	void setStartStates(const vector<int> &startStates);
+	void generateSequenceOfMoves(const vector<int> &endStates, vector<Move> &sequenceOfMoves);
+
+private:
+	int findEmptySlotId();
+	vector<int> findCarPositions(const vector<int> &currentStates);
+	int findMisplacedCar(const vector<int> &carPosition, const vector<int> &endStates);
+	void RearrangeCars::moveCar(vector<int> &currentStates, vector<int> &carPosition, int from, int to);
```
July-Morning: I suppose there should be just moveCar instead of RearrangeCars::moveCar

group/assignment6/Move.h
```diff
@@ -0,0 +1,20 @@
+#ifndef MOVE_H
+#define MOVE_H
+#pragma once
+
+struct Move
+{
+	int carId;
+	int startPosition;
```
July-Morning: I still believe that startPosition and endPosition should be treated as a pair of ints so that they would be separated from the carId. What do you think?

group/assignment6/TestRearrangeCars.cpp
```diff
@@ -0,0 +1,30 @@
...
+}
+
+template<typename T>
+void printSequence(vector<T> sequence){
+	for (T &item : sequence)
+		cout << item;
+}
+
+void TestRearrangeCars::testValidRearrangingOfCars() {
+	vector<int> startStates = { 0, 1, -1, 2 };
```
July-Morning: I started to draft some test cases, and for in case of empty startStates and/or endStates vectors the code fails, I suppose there should be some check for that. Same for the case when startStates and endStates have different lengths. Same when there are several empty slots in startStates and/or endStates, same when there are duplicates in startStates and/or endStates. Of course we can fix it later, when all the tests will be ready. Maybe we should add some additional method to valudate input vectors?

group/assignment6/RearrangeCars.cpp
```diff
@@ -0,0 +1,56 @@
...
+	return -1; // If every car is on its slot
+}
+
+void RearrangeCars::moveCar(vector<int> &currentStates, vector<int> &carPosition, int from, int to){
+	currentStates[to] = currentStates[from];
+	currentStates[from] = -1;
+	carPosition[currentStates[to]] = to;
+}
+
+void RearrangeCars::generateSequenceOfMoves(const vector<int> &endStates, vector<Move> &sequenceOfMoves){
```
July-Morning: Maybe bool should be returned instead of void (false of the invalid input case)?

group/assignment6/RearrangeCars.cpp
```diff
@@ -0,0 +1,56 @@
...
+	for (int slotId = 0; slotId < this->startStates.size(); slotId++){
+		if (this->startStates[slotId] == -1) return slotId; // no car has -1 id, it is reserved for empty slot
+	}
+	return -1; // This code should not be reached
+}
+
+vector<int> RearrangeCars::findCarPositions(const vector<int> &currentStates){
+	vector<int> carPositions(currentStates.size() - 1);
+	for (int i = 0; i < currentStates.size(); i++)
+	{
```
July-Morning: You always keep { on the same lane as for, let's follow this style. Btw {} are not really necessary for this for loop (same in some other places in the code, we can remove them).

group/assignment6/RearrangeCars.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+void RearrangeCars::generateSequenceOfMoves(const vector<int> &endStates, vector<Move> &sequenceOfMoves){
+	vector<int> currentStates = this->startStates; // startStates will remain unchanged
+	vector<int> carPosition = findCarPositions(currentStates);
+	
+	int emptySlotId = findEmptySlotId();
+	int currCarId, currCarSlotId;
+	
+	while (true){
+		currCarId = endStates[emptySlotId];
```
July-Morning: Maybe here we should better put 
`if (endStates[emptySlotId] == -1) // We reached slot which should be empty in the end
			currCarId = findMisplacedCar(carPosition, endStates);`
Otherwise it seems we use currCarId for two different purposes, I believe...

group/assignment6/RearrangeCars.cpp
TijanaSonjic: I agree, so method which validates input vectors will be called at the beginning of this function, not in a test before calling this function?
I think that calling that method here will be a better option.

July-Morning: I think so. We shouldn't let user think about validation, better to do it here for certain.

group/assignment6/RearrangeCars.cpp
```diff
@@ -0,0 +1,56 @@
...
+}
+
+void RearrangeCars::moveCar(vector<int> &currentStates, vector<int> &carPosition, int from, int to){
+	currentStates[to] = currentStates[from];
+	currentStates[from] = -1;
+	carPosition[currentStates[to]] = to;
+}
+
+void RearrangeCars::generateSequenceOfMoves(const vector<int> &endStates, vector<Move> &sequenceOfMoves){
+	vector<int> currentStates = this->startStates; // startStates will remain unchanged
```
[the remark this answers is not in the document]
natiami: right now you can't do that. Also I don't think it will make solution easier. I mean it saves startStates and every time you call  generateSequenceOfMoves you pass different endstates (I don't know why someone would calculate the same twice). But :) if you do think otherwise please do so, I don't mind

group/assignment6/Move.h
natiami: I don't mind, so have a pair in structure? 

struct Move
{
	int carId;
	pair<int, int> positions;
This way? or you prefer vector of pairs of pairs without structure?

group/assignment6/TestRearrangeCars.cpp
natiami: I agree. We can add methods for that

group/assignment6/Move.h
July-Morning: Yes, this way is good for me, the structure is needed to keep code readable.

PR #40, opened 2018-07-30
Add bruteForce solution (draft)

annnufan:
I'll make it better at the evening 
But main idea clear right now

### DIFF
--- a/group/assignment6/RearrangeCars.cpp
+++ b/group/assignment6/RearrangeCars.cpp
@@ -1,0 +1,172 @@
+#include "RearrangeCars.h"
+
+RearrangeCars::RearrangeCars(const vector<int> &startStates) : startStates(startStates) {}
+
+void RearrangeCars::setStartStates(const vector<int> &startStates) {
+	this->startStates = startStates;
+}
+
+int RearrangeCars::findEmptySlotId() {
+	for (int slotId = 0; slotId < this->startStates.size(); slotId++) {
+		if (this->startStates[slotId] == -1) return slotId; // no car has -1 id, it is reserved for empty slot
+	}
+	return -1; // This code should not be reached
+}
+
+vector<int> RearrangeCars::findCarPositions(const vector<int> &currentStates) {
+	vector<int> carPositions(currentStates.size() - 1);
+	for (int i = 0; i < currentStates.size(); i++) {
+		if (currentStates[i] != -1) // check for empty slot
+			carPositions[currentStates[i]] = i;
+	}
+	return carPositions;
+}
+
+int RearrangeCars::findMisplacedCar(const vector<int> &carPosition, const vector<int> &endStates) {
+	for (int carId = 0; carId < carPosition.size(); carId++) {
+		if (endStates[carPosition[carId]] != carId) return carId;
+	}
+	return -1; // If every car is on its slot
+}
+
+void RearrangeCars::moveCar(vector<int> &currentStates, vector<int> &carPosition, int from, int to) {
+	currentStates[to] = currentStates[from];
+	currentStates[from] = -1;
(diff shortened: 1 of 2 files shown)


group/assignment6/RearrangeCars.cpp
```diff
@@ -139,3 +139,34 @@ void RearrangeCars::generateSequenceOfMovesWithConstraints(const vector<int> &en
...
 	}
 }
+
+void RearrangeCars::bruteForceAllSequence(vector<int>& currentStates, const vector<int>& endStates, vector<int>& carPosition, int emptySlotId, vector<Move>& sequenceOfMoves, vector<vector<Move>>& sequencesOfMoves, set<vector<int>>& usedPositions) {
+	int numOfCarsOnDisiredPosition = numberOfCarsOnDesiredPosition(carPosition, endStates);
+	if (numOfCarsOnDisiredPosition == carPosition.size()) {
+		sequencesOfMoves.push_back(sequenceOfMoves);
+		return;
+	}
+	for (int currCarSlotId = 0; currCarSlotId < carPosition.size(); currCarSlotId++) {
```
michbl: I think there is a little problem here - there are N slots and the size of carPosition is N - 1. That means we will never get to the last slot. I guess we can change that to _currCarSlotID < currentStates.size();_

michbl: I think there is a little problem here - there are N slots and the size of carPosition is N - 1. That means we will never get to the last slot. I guess we can change that to _currCarSlotID < currentStates.size();_

group/assignment6/RearrangeCars.cpp
```diff
@@ -139,3 +139,34 @@ void RearrangeCars::generateSequenceOfMovesWithConstraints(const vector<int> &en
...
+		if (currCarSlotId != emptySlotId) {
```
michbl: I think we should add another check here - _if (currCarSlotId != emptySlotId && validMove(carId, emptySlotId, constraints))_, and then this solution also supports constraints

group/assignment6/RearrangeCars.cpp
```diff
@@ -139,3 +139,34 @@ void RearrangeCars::generateSequenceOfMovesWithConstraints(const vector<int> &en
...
+			moveCar(currentStates, carPosition, currCarSlotId, emptySlotId);
+			if (usedPositions.find(carPosition) == usedPositions.end()) {
+				usedPositions.insert(carPosition);
+				sequenceOfMoves.push_back(Move(currentStates[emptySlotId], make_pair(currCarSlotId, emptySlotId)));
+				bruteForceAllSequence(currentStates, endStates, carPosition, currCarSlotId, sequenceOfMoves, sequencesOfMoves, usedPositions);
+				sequenceOfMoves.pop_back();
```
michbl: This section was confusing to me - I understand that we swapped the car in _currCarSlotId_ with the "car" in _emptySlotId_, but it's really hard to understand that the values in _currentStates_ changed when you first read it. Maybe we can also swap the values in _currCarSlotId_ and _emptySlotId_ after we push the move to the sequence of moves, or at least add a comment that explains what's happening here. What do you think?

group/assignment6/RearrangeCars.cpp
annnufan: I'll check it right now. But as I think, we need to try all N positions

group/assignment6/RearrangeCars.cpp
annnufan: Yeah, it's easy and add functionality to solution

annnufan:
About name.. I'm not sure/ Now I just try to use consistent name for clearly reading

PR #42, opened 2018-08-02
Group 6 iteration 1

TijanaSonjic:
Merged basic methods and methods that work with constrains. Last parameter of those functions is now optional. @July-Morning , I didn't delete tests for basic solution just to not disrupt any use case. Please do this part :) 
Girls, do the review only for this pull request. Next iterations will be over this code.

### DIFF
--- a/group/assignment6/RearrangeCars.cpp
+++ b/group/assignment6/RearrangeCars.cpp
@@ -44,7 +44,7 @@ bool RearrangeCars::validMove(int carId, int emptySlotId, const vector<set<int>>
 	return false;
 }
 
-int RearrangeCars::numberOfCarsOnDesiredPosition(const vector<int>& carPosition, const vector<int>& endStates) {
+int RearrangeCars::numberOfCarsOnTheirDesiredPositions (const vector<int>& carPosition, const vector<int>& endStates) {
 	int count = 0;
 	for (int i = 0; i < carPosition.size(); i++)
 		if (endStates[carPosition[i]] == i)
@@ -102,7 +102,7 @@ void RearrangeCars::generateSequenceOfMoves(const vector<int> &endStates, vector
 
 	int emptySlotId = findEmptySlotId();
 	int currCarId, currCarSlotId;
-	int numOfCarsOnDisiredPosition = numberOfCarsOnDesiredPosition(carPosition, endStates);
+	int numOfCarsOnDisiredPosition = numberOfCarsOnTheirDesiredPositions (carPosition, endStates);
 
 	while (true) {
 		currCarId = endStates[emptySlotId];
@@ -121,7 +121,7 @@ void RearrangeCars::generateSequenceOfMoves(const vector<int> &endStates, vector
 }
 
 void RearrangeCars::bruteForceAllSequence(vector<int>& currentStates, const vector<int>& endStates, vector<int>& carPosition, int emptySlotId, vector<Move>& sequenceOfMoves, vector<vector<Move>>& sequencesOfMoves, set<vector<int>>& usedPositions, const vector<set<int>> *constraints) {
-	int numOfCarsOnDisiredPosition = numberOfCarsOnDesiredPosition(carPosition, endStates);
+	int numOfCarsOnDisiredPosition = numberOfCarsOnTheirDesiredPositions (carPosition, endStates);
 	if (numOfCarsOnDisiredPosition == carPosition.size()) {
 		sequencesOfMoves.push_back(sequenceOfMoves);
 		return;
--- a/group/assignment6/RearrangeCars.cpp
+++ b/group/assignment6/RearrangeCars.cpp
@@ -8,17 +8,20 @@ void RearrangeCars::setStartStates(const vector<int> &startStates) {
 
 int RearrangeCars::findEmptySlotId() {
 	for (int slotId = 0; slotId < this->startStates.size(); slotId++) {
-		if (this->startStates[slotId] == -1) return slotId; // no car has -1 id, it is reserved for empty slot
+		if (this->startStates[slotId] == -1) 
+			return slotId;		// no car has -1 id, it is reserved for empty slot
 	}
-	return -1; // This code should not be reached
+
+	return -1;	// This code should not be reached
 }
 
 vector<int> RearrangeCars::findCarPositions(const vector<int> &currentStates) {
 	vector<int> carPositions(currentStates.size() - 1);
 	for (int i = 0; i < currentStates.size(); i++) {
-		if (currentStates[i] != -1) // check for empty slot
+		if (currentStates[i] != -1)		// check for empty slot
 			carPositions[currentStates[i]] = i;
 	}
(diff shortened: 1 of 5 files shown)


group/assignment6/RearrangeCars.cpp
```diff
@@ -35,36 +31,29 @@ void RearrangeCars::moveCar(vector<int> &currentStates, vector<int> &carPosition
...
-		if (currCarId == -1)	// We reached slot which should be empty in the end
-			currCarId = findMisplacedCar(carPosition, endStates);
-		if (currCarId == -1) break;
-		currCarSlotId = carPosition[currCarId];
-		moveCar(currentStates, carPosition, currCarSlotId, emptySlotId);
-		sequenceOfMoves.push_back(Move(currCarId, make_pair(currCarSlotId, emptySlotId)));
-		emptySlotId = currCarSlotId;
-	}
-}
+bool RearrangeCars::validMove(int carId, int emptySlotId, const vector<set<int>>* constrains) {
```
natiami: function name should be isMoveValid I think :) for me it's more understandable

group/assignment6/RearrangeCars.cpp
```diff
@@ -35,36 +31,29 @@ void RearrangeCars::moveCar(vector<int> &currentStates, vector<int> &carPosition
...
 
-	if ((constrains[emptySlotId]).find(carId) == (constrains[emptySlotId].end()))
+	if (((*constrains)[emptySlotId]).find(carId) == ((*constrains)[emptySlotId].end()))
 		return true;
 
 	return false;
 }
 
-int RearrangeCars::findMisplacedCarWithConstraints(vector<int> &carPosition, vector<int> &currentStates, const vector<int> &endStates, const vector<set<int>> &constraints, vector<Move> &sequenceOfMoves, int &emptySlotId, int &numOfCarsOnDisiredPosition) {
+int RearrangeCars::numberOfCarsOnDesiredPosition(const vector<int>& carPosition, const vector<int>& endStates) {
```
natiami: placedCarsAmount, numberOfPlacedCars, I don't like current name because it should be numberOfCarsOnTheirDesiredPositions otherwise it kind of sounds that we check car numbers on one desired position. But if you don't agree leave it :)

group/assignment6/RearrangeCars.cpp
```diff
@@ -35,36 +31,29 @@ void RearrangeCars::moveCar(vector<int> &currentStates, vector<int> &carPosition
...
+	int count = 0;
+	for (int i = 0; i < carPosition.size(); i++)
+		if (endStates[carPosition[i]] == i)
+			count++;
+
+	return count;
+}
+
+int RearrangeCars::findMisplacedCar(vector<int> &carPosition, vector<int> &currentStates, const vector<int> &endStates, const vector<set<int>> *constraints, vector<Move> &sequenceOfMoves, int &emptySlotId, int &numOfCarsOnDisiredPosition) {
```
natiami: when you have too many parameters put them under each other to make it easier for reader. But then I searched what is appropriate way to pass many parameters and found out that we should make a structure struct. For example, parkingLot and it will have fields: carPosition, currentStates, endStates, constraints, emptySlotId and numOfCarsOnDisiredPosition. so we pass only that structure object and sequenceOfMoves to function. bruteForceAllSequence function can also use this structure. What you think?

natiami: Also this function does not only finds misplaced car, so name must be changed.

group/assignment6/RearrangeCars.cpp
```diff
@@ -127,7 +107,7 @@ void RearrangeCars::generateSequenceOfMovesWithConstraints(const vector<int> &en
 	while (true) {
```
natiami: numberOfCarsOnDesiredPosition why call this function? how can it be something other than 0 here? I don't get the idea :) please explain

group/assignment6/RearrangeCars.cpp
```diff
@@ -140,33 +120,38 @@ void RearrangeCars::generateSequenceOfMovesWithConstraints(const vector<int> &en
 	}
 }
 
-void RearrangeCars::bruteForceAllSequence(vector<int>& currentStates, const vector<int>& endStates, const vector<set<int>> &constraints, vector<int>& carPosition, int emptySlotId, vector<Move>& sequenceOfMoves, vector<vector<Move>>& sequencesOfMoves, set<vector<int>>& usedPositions) {
+void RearrangeCars::bruteForceAllSequence(vector<int>& currentStates, const vector<int>& endStates, vector<int>& carPosition, int emptySlotId, vector<Move>& sequenceOfMoves, vector<vector<Move>>& sequencesOfMoves, set<vector<int>>& usedPositions, const vector<set<int>> *constraints) {
```
natiami: I don't like the function name bruteForceAllSequence :) call it generateMoves or findAllSequencesOfMoves

July-Morning:
No, I need to fix them according to the changes. Could you please merge this iteration @TijanaSonjic ?

July-Morning:
@natiami  I agree with nearly all your comments, but I think we should now first put everything together and then do a proper clean-up...

group/assignment6/TestRearrangeCars.cpp
```diff
@@ -15,41 +15,41 @@ void TestRearrangeCars::runAllTests() {
 }
 
 void TestRearrangeCars::testNoConstraints() {
```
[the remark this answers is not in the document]
July-Morning: I will fix the tests tomorrow. Names will be simplier as now we do not separate basic case and case with constraints. Not sure if comments are better than long names (not that long maybe), comments are more for "why" than "what"...

group/assignment6/RearrangeCars.cpp
July-Morning: I agree here.

group/assignment6/RearrangeCars.cpp
July-Morning: I think this can be postponed to the next iteration so that we can further discuss if we need a struct.

group/assignment6/RearrangeCars.cpp
July-Morning: Agree.

group/assignment6/RearrangeCars.cpp
TijanaSonjic: If endPosition is the same as startPosition for any car. When number of cars on desired position equals total number of cars, we know that we have to finish an algorithm.

group/assignment6/RearrangeCars.cpp
TijanaSonjic: Me too

group/assignment6/RearrangeCars.cpp
TijanaSonjic: It makes sense, I will rename it.

TijanaSonjic:
Sure, I am going to merge it now, so we can continue iterating over branch6.

PR #43, opened 2018-08-03
Tests iteration 1

July-Morning:
I fixed the tests according to the iteration 1 changes in the code and added some tests for challenge #4. I did not duplicate tests for invalid input, because I think it would be better to separate them later as test for input validation method that we are going to create. Do you agree?


July-Morning:
Thanks! I'm merging then.

PR #44, opened 2018-08-03
Added input validity check

michbl:
Hi girls,

I added the function _bool isInputValid()_. I tried to document it as clearly as I could.
I added input validity checks in the functions _generateSequenceOfMoves_, _generateAllSequencesOfMoves_.
I also added another thing in each of these functions: they both recieve a vector of moves from the user that we are supposed to fill, so to make sure that they are empty I cleared these vectors at the beginning of the functions.
Last change: I changed _generateSequenceOfMoves_, _generateAllSequencesOfMoves_ to return bool, because if the input isn't valid the user has no indication of that. We can throw exceptions instead if you want..
Tell me what you think :)

And most of the changes (tabs) are because of my IDE.. :((

### DIFF
--- a/group/assignment6/RearrangeCars.cpp
+++ b/group/assignment6/RearrangeCars.cpp
@@ -181,7 +181,7 @@ bool RearrangeCars::generateAllSequencesOfMoves(const vector<int> &endStates, ve
 }
 
 bool RearrangeCars::isInputValid(const vector<int> &endStates, const vector<set<int>> *constraints) {
-      bool atLeastOneCar = !endStates.size() >= 2 && !this->startStates.size() >= 2;
+      bool atLeastOneCar = endStates.size() >= 2 && this->startStates.size() >= 2;
       bool stateVectorsHaveSameSize = endStates.size() == this->startStates.size();
       bool startStatesValid = isInputValidAux(this->startStates, constraints);
       bool endStatesValid = isInputValidAux(endStates, constraints);
--- a/group/assignment6/RearrangeCars.cpp
+++ b/group/assignment6/RearrangeCars.cpp
@@ -181,12 +181,12 @@ bool RearrangeCars::generateAllSequencesOfMoves(const vector<int> &endStates, ve
 }
 
 bool RearrangeCars::isInputValid(const vector<int> &endStates, const vector<set<int>> *constraints) {
-      bool stateVectorsNotEmpty = !endStates.empty() && !this->startStates.empty();
+      bool atLeastOneCar = !endStates.size() >= 2 && !this->startStates.size() >= 2;
       bool stateVectorsHaveSameSize = endStates.size() == this->startStates.size();
       bool startStatesValid = isInputValidAux(this->startStates, constraints);
       bool endStatesValid = isInputValidAux(endStates, constraints);
 
-      return stateVectorsNotEmpty && stateVectorsHaveSameSize && startStatesValid && endStatesValid;
+      return atLeastOneCar && stateVectorsHaveSameSize && startStatesValid && endStatesValid;
 }
 
 bool RearrangeCars::isInputValidAux(const vector<int> &states, const vector<set<int>> *constraints) {
@@ -195,7 +195,7 @@ bool RearrangeCars::isInputValidAux(const vector<int> &states, const vector<set<
       for (int slotId = 0; slotId < states.size(); slotId++) {
             int carId = states[slotId];
 
-            // make sure there is exactly one empty slot
+            // make sure there is not more than one empty slot
             if (carId == EMPTY_SLOT_ID) {
                   if (foundEmptySlot) {
                         return false;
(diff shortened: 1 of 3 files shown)


group/assignment6/RearrangeCars.cpp
```diff
@@ -3,155 +3,218 @@
...
+      currentStates[to] = currentStates[from];
+      currentStates[from] = -1;
+      carPosition[currentStates[to]] = to;
 }
 
-bool RearrangeCars::validMove(int carId, int emptySlotId, const vector<set<int>>* constrains) {
-	if (!constrains)	// Constrains are not set
-		return true;
+bool RearrangeCars::validMove(int carId, int emptySlotId, const vector<set<int>> *constrains) {
+      if (!constrains)        // Constrains are not set
```
July-Morning: I think, these are constrainTs, not constrains - we should fix this typo everywhere.

group/assignment6/RearrangeCars.cpp
```diff
@@ -3,155 +3,218 @@
...
+      }
 }
 
-void RearrangeCars::generateAllSequencesOfMoves(const vector<int> &endStates, vector<vector<Move>> &sequencesOfMoves, const vector<set<int>> *constraints) {
-	vector<int> currentStates = this->startStates;
-	vector<int> carPosition = findCarPositions(currentStates);
+bool RearrangeCars::generateAllSequencesOfMoves(const vector<int> &endStates, vector<vector<Move>> &sequencesOfMoves,
+                                                const vector<set<int>> *constraints) {
+      sequencesOfMoves.clear();
+      if (!isInputValid(this->startStates) || !isInputValid(endStates)) {
```
July-Morning: I believe it is a bit overcomplicated, don't you think so? I don't really understand why should we call isInputValid several times and why don't we put constraints as the argument of input validation. Could you explain this for me, please? I thought it would be something like
`if (!isInputValid(endStates, constraints)) 
    return false;`

group/assignment6/RearrangeCars.h
```diff
@@ -6,27 +6,36 @@
...
+      int findMisplacedCar(vector<int> &carPosition, vector<int> &currentStates, const vector<int> &endStates,
+                           const vector<set<int>> *constraints, vector<Move> &sequenceOfMoves, int &emptySlotId,
+                           int &numOfCarsOnDisiredPosition);
+      void moveCar(vector<int> &currentStates, vector<int> &carPosition, int from, int to);
+      bool validMove(int carId, int emptySlot, const vector<set<int>> *constraints);
+      int numberOfCarsOnTheirDesiredPositions(const vector<int> &carPosition, const vector<int> &endPosition);
+      void bruteForceAllSequence(vector<int> &currentStates, const vector<int> &endStates, vector<int> &carPosition,
+                                 int emptySlotId, vector<Move> &sequenceOfMoves, vector<vector<Move>> &sequencesOfMoves,
+                                 set<vector<int>> &usedPositions, const vector<set<int>> *constraints = nullptr);
+      bool isInputValid(const vector<int> &startStates, bool endStates = false,
```
July-Morning: Why does this method have 3 arguments, but only 1 is used in our solution?

July-Morning:
Oh, I was looking to the outdated version before @TijanaSonjic  fixed the breaks. But I beleive my comments are still valid.

July-Morning:
Btw I can confirm that challenge #4 is working as it should now (at least on the tests we have),  thanks for fixing that, @michbl ! And returning bool is good to me, I don't think we need exceptions here. Clearing vectors is also a very reasonable change! :+1:

group/assignment6/RearrangeCars.cpp
TijanaSonjic: I will correct this :)

group/assignment6/RearrangeCars.cpp
michbl: You're absolutely right, my bad
Fixing

group/assignment6/RearrangeCars.h
michbl: I will fix that
It was really late at night when I wrote this :#

July-Morning: Be sure I understand this pretty well :)

July-Morning:
Now it's much better to me! Few more things to fix:
1) We must have one and only one empty slot, so there's a need an extra check if `foundEmptySlot` is true after the cycle.
2) Having one empty slot  { -1 } as the startStates and as the endStates with empty constraints seems to be a valid case, but we get a segmentation error here. Same error with the invalid constraints - but this must be caught by isInputValid I guess.
3) Still some concerns about challenge #4: as I understand it does not take canstraints into account, so we should fix that or remove this parameter. 

What do you think? I am a bit busy now but I can debug this in the evening.

michbl:
@July-Morning I will fix 1 and 2 but about 3 - we are using constraints in line 145: _if (currCarSlotId != emptySlotId && validMove(currentStates[currCarSlotId], emptySlotId, constraints)) {_
Isn't that enough?

July-Morning:
I thought one empty slot case is valid, but ok, we can say that we must have at least one car. @michbl could you please check that this version works fine and all tests pass? If yes, I think it's time to merge and start review process :)

michbl:
All the tests passed, merging

PR #45, opened 2018-08-04
Working solution

michbl:
This is our latest working solution
Everyone is welcome to review 😀

### DIFF
--- a/group/assignment6/RearrangeCars.cpp
+++ b/group/assignment6/RearrangeCars.cpp
@@ -108,10 +108,10 @@ int RearrangeCars::findMisplacedCarWithConstraints(vector<int> &carPosition, vec
 }
 
 int RearrangeCars::numberOfCarsOnDesiredPosition(const vector<int>& carPosition, const vector<int>& endStates) {
-	int num = 0;
+	int count = 0;
 	for (int i = 0; i < carPosition.size(); i++)
 		if (endStates[carPosition[i]] == i)
-			num++;
+			count++;
 
 	return num;
 }
--- a/group/assignment6/RearrangeCars.cpp
+++ b/group/assignment6/RearrangeCars.cpp
@@ -1,0 +1,172 @@
+#include "RearrangeCars.h"
+
+RearrangeCars::RearrangeCars(const vector<int> &startStates) : startStates(startStates) {}
+
+void RearrangeCars::setStartStates(const vector<int> &startStates) {
+	this->startStates = startStates;
+}
+
+int RearrangeCars::findEmptySlotId() {
+	for (int slotId = 0; slotId < this->startStates.size(); slotId++) {
+		if (this->startStates[slotId] == -1) return slotId; // no car has -1 id, it is reserved for empty slot
+	}
+	return -1; // This code should not be reached
+}
+
+vector<int> RearrangeCars::findCarPositions(const vector<int> &currentStates) {
+	vector<int> carPositions(currentStates.size() - 1);
+	for (int i = 0; i < currentStates.size(); i++) {
+		if (currentStates[i] != -1) // check for empty slot
+			carPositions[currentStates[i]] = i;
+	}
+	return carPositions;
+}
+
+int RearrangeCars::findMisplacedCar(const vector<int> &carPosition, const vector<int> &endStates) {
+	for (int carId = 0; carId < carPosition.size(); carId++) {
+		if (endStates[carPosition[carId]] != carId) return carId;
+	}
+	return -1; // If every car is on its slot
+}
+
+void RearrangeCars::moveCar(vector<int> &currentStates, vector<int> &carPosition, int from, int to) {
+	currentStates[to] = currentStates[from];
+	currentStates[from] = -1;
+	carPosition[currentStates[to]] = to;
+}
+
+void RearrangeCars::generateSequenceOfMoves(const vector<int> &endStates, vector<Move> &sequenceOfMoves) {
+	vector<int> currentStates = this->startStates; // startStates will remain unchanged
+	vector<int> carPosition = findCarPositions(currentStates);
+
+	int emptySlotId = findEmptySlotId();
+	int currCarId, currCarSlotId;
+
+	while (true) {
+		currCarId = endStates[emptySlotId];
+		if (currCarId == -1)	// We reached slot which should be empty in the end
+			currCarId = findMisplacedCar(carPosition, endStates);
+		if (currCarId == -1) break;
+		currCarSlotId = carPosition[currCarId];
+		moveCar(currentStates, carPosition, currCarSlotId, emptySlotId);
+		sequenceOfMoves.push_back(Move(currCarId, make_pair(currCarSlotId, emptySlotId)));
+		emptySlotId = currCarSlotId;
+	}
+}
+
+bool RearrangeCars::validMove(int carId, int emptySlotId, const vector<set<int>>& constrains) {
+	if (constrains[emptySlotId].size() == 0)	// There are no constraints for this parking space, any car can be moved here
+		return true;
+
+	if ((constrains[emptySlotId]).find(carId) != (constrains[emptySlotId].end()))
+		return true;
+
+	return false;
+}
+
+int RearrangeCars::findMisplacedCarWithConstraints(vector<int> &carPosition, vector<int> &currentStates, const vector<int> &endStates, const vector<set<int>> &constraints, vector<Move> &sequenceOfMoves, int &emptySlotId, int &numOfCarsOnDisiredPosition) {
+	int lastMovedCarId = -1;
+	int currCarSlotId;
+
+	while (true) {
+		currCarSlotId = endStates[emptySlotId];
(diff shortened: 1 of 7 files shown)


group/assignment6/RearrangeCars.h
```diff
@@ -0,0 +1,43 @@
...
+#include "Move.h"
+#include <iostream>
+#include <vector>
+#include <queue>
+#include <set>
+#include <unordered_map>
+
+#define EMPTY_SLOT_ID -1
+
+using namespace std;
```
TijanaSonjic: We do not need to import this namespace, this line should be removed.

group/assignment6/RearrangeCars.cpp
```diff
@@ -0,0 +1,221 @@
...
+RearrangeCars::RearrangeCars(const vector<int> &startStates) : startStates(startStates) {}
+
+void RearrangeCars::setStartStates(const vector<int> &startStates) {
+      this->startStates = startStates;
+}
+
+int RearrangeCars::findEmptySlotId() {
+      for (int slotId = 0; slotId < this->startStates.size(); slotId++) {
+            if (this->startStates[slotId] == EMPTY_SLOT_ID)
+                  return slotId;                // no car has EMPTY_SLOT_ID id, it is reserved for empty slot
```
TijanaSonjic: I find our that lines that are non-obvious should get a comment at the end of the line. These end-of-line comments should be separated from the code by 2 spaces.

group/assignment6/RearrangeCars.cpp
```diff
@@ -0,0 +1,221 @@
...
+            emptySlotId = currCarSlotId;
+
+            if (numOfCarsOnDisiredPosition == carPosition.size())
+                  break;
+      }
+      return true;
+}
+
+void
+RearrangeCars::bruteForceAllSequence(vector<int> &currentStates, const vector<int> &endStates, vector<int> &carPosition,
```
TijanaSonjic: Must be renamed!

TijanaSonjic (commented):
Girls, I like our solution. There are few things I think that has to be changed.  I will update this pull request with  formatted code, comments improvement according to style guide,.. because I don't know what will be status of my internet connection.

group/assignment6/RearrangeCars.h
```diff
@@ -0,0 +1,43 @@
...
+private:
+      int findEmptySlotId();
+      vector<int> findCarPositions(const vector<int> &currentStates);
+      int findMisplacedCar(vector<int> &carPosition, vector<int> &currentStates, const vector<int> &endStates,
+                           const vector<set<int>> *constraints, vector<Move> &sequenceOfMoves, int &emptySlotId,
+                           int &numOfCarsOnDisiredPosition);
+      void moveCar(vector<int> &currentStates, vector<int> &carPosition, int from, int to);
+      bool validMove(int carId, int emptySlot, const vector<set<int>> *constraints);
+      int numberOfCarsOnTheirDesiredPositions(const vector<int> &carPosition, const vector<int> &endPosition);
+      void bruteForceAllSequence(vector<int> &currentStates, const vector<int> &endStates, vector<int> &carPosition,
```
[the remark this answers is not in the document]
July-Morning: I believe we should leave it splitted. Maybe a bit better solution would be a struct for the set of arguments, but it seems we don't have time to do that.

group/assignment6/TestRearrangeCars.cpp
```diff
@@ -0,0 +1,281 @@
...
+template<typename T>
+void printSequence(vector<T> sequence) {
+	for (T &item : sequence)
+		cout << item;
+	cout << endl;
+}
+
+template<typename T>
+void printAllSequences(vector<vector<T>> allSequences) {
+	for (vector<T> &sequence : allSequences)
```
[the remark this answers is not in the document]
July-Morning: Thanks, my typo!

group/assignment6/RearrangeCars.cpp
July-Morning: I agree. Also sequenceOfMoves and sequencesOfMoves are very similar to each other so it's easy to get confused here. Maybe we should rename arguments, too?

In my opinion method might be called "getAllSequencesFromCurPosition", and arguments might be "sequenceOfMoves" and "allSequencesOfMoves", what do you think?

group/assignment6/RearrangeCars.cpp
```diff
@@ -0,0 +1,227 @@
...
+}
+
+void RearrangeCars::moveCar(vector<int> &currentStates, vector<int> &carPosition, int from, int to) {
+	currentStates[to] = currentStates[from];
+	currentStates[from] = EMPTY_SLOT_ID;
+	carPosition[currentStates[to]] = to;
+}
+
+bool RearrangeCars::validMove(int carId, int emptySlotId, const vector<set<int>> *constraints) {
+	if (!constraints)  // Constraints are not set.
```
July-Morning: Sometimes we put brackets around one-string if-s and for-s, sometimes we don't. I think one style should be chosen.

group/assignment6/RearrangeCars.cpp
michbl: I like it. We can also call it "getAllSequencesFromCur*rent*Position" if we're using a long name anyways..
I also agree about the variable names

Also about variable names, I said it before but nobody said what she thinks.. I think startStates, endStates, carPosition are very confusing name.. I would call them currentSlot2CarId, desiredSlot2CarId, currentCarId2Slot. Or something like that.. What do you think?

July-Morning (commented):
Our solution works, and it seems mostly all of us are now travelling or packing staff for the journey :) Thus, I believe we should do our best to clean the code, but we shouldn't make any significant changes, because there's no time to discuss and test them.

group/assignment6/RearrangeCars.cpp
July-Morning: Oh yes, I just wanted to comment about Current. Sometimes we shorten it to Curr,  sometimes we use the full word. I think both are ok, but my mistake here was to shorten it to Cur.

Well... For me startStates, endStates and carPositions are pretty understandable, but maybe it's because I got used to them during this assignment. I think your version is better :+1:  If everyone agrees, we can fix that.

group/assignment6/RearrangeCars.cpp
```diff
@@ -0,0 +1,227 @@
...
+      
+	return true;
+}
+
+void
+RearrangeCars::bruteForceAllSequence(vector<int> &currentStates, const vector<int> &endStates, vector<int> &carPosition,
+				      int emptySlotId, vector<Move> &sequenceOfMoves,
+			              vector<vector<Move>> &sequencesOfMoves, set<vector<int>> &usedPositions,
+				      const vector<set<int>> *constraints) {
+	int numOfCarsOnDisiredPosition = numberOfCarsOnTheirDesiredPositions(carPosition, endStates);
```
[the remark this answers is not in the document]
michbl: You're right

group/assignment6/RearrangeCars.cpp
```diff
@@ -0,0 +1,227 @@
...
+	vector<Move> sequenceOfMoves;
+	set<vector<int>> usedPositions;
+	usedPositions.insert(carPosition);
+	bruteForceAllSequence(currentStates, endStates, carPosition, emptySlotId, sequenceOfMoves, sequencesOfMoves, usedPositions, constraints);
+	
+      return true;
+}
+
+bool RearrangeCars::isInputValid(const vector<int> &endStates, const vector<set<int>> *constraints) {
+	bool atLeastOneCar = endStates.size() >= 2 && this->startStates.size() >= 2;
```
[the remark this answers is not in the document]
michbl: I do

group/assignment6/TestRearrangeCars.cpp
```diff
@@ -0,0 +1,280 @@
...
+	vector<int> startStates = { 0, 1, -1, 2 };
+	vector<int> endStates = { 2, -1, 1, 0 };
+
+	testValidInputAllSequences(startStates, endStates);
+}
+
+void TestRearrangeCars::testValidStatesAllSequences2() {
+	vector<int> startStates = { 0, 1, -1 };
+	vector<int> endStates = { 1, 0, -1 };
+
```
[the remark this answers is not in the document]
michbl: I can try to see why it doesn't work but if it takes too much time I will remove it. Is that ok?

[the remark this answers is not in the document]
July-Morning: Sure! Now I again think that this might be my mistake as I was not passing constraints in testValidInputAllSequences... Was in the hurry, sorry. But I will be able to check this only few hours later.

group/assignment6/RearrangeCars.cpp
July-Morning: I fixed startStates and endStates names, but kept carPositions. For me introducing CarId2Slot may be a bit confusing for thre reader, but I am ready to discuss.

July-Morning:
I fixed as much as I could according to the comments by @TijanaSonjic  and @michbl. As our deadline is close, it seems that current version is more or less our final. Please review it if you can!

July-Morning (approved):

